PR #29, opened 2020-08-27
Add zstd compression when fountain encoding

sz3:
Includes a minimal source tree of [zstd](https://github.com/facebook/zstd/tree/dev/lib) under `third_party_lib/zstd`.

Compression is enabled only for fountain encoding (for now), and on by default. Compression level is set to `6` by default, but configurable.

This PR includes a workaround for the issue of data smaller than "750" bytes being rejected by the fountain encoder. Specifically, the issue is:
* wirehair will not work on data smaller than a single frame
* frame size is calculated by cimbar as `post-error-correction bytes / 10`
   * `750` for `ecc=30`, 6-bit cimbar (the default)
   * `980` for `ecc=15`, 7-bit cimbar (realistic high-end)

I say *workaround*, because the current method is to use a [zstd skippable frame](https://tools.ietf.org/html/rfc8478#section-3.1.2) to pad the data with 0s until it is larger than a single wirehair frame. :tada: 
     * this is, clearly, silly
     * but it works!


src/lib/encoder/Encoder.h
```diff
@@ -23,8 +24,8 @@ class Encoder
 	std::optional<cv::Mat> encode_next(STREAM& stream);
 
 	unsigned encode(const std::string& filename, std::string output_prefix);
-	unsigned encode_fountain(const std::string& filename, std::string output_prefix);
-	unsigned encode_fountain(const std::string& filename, const std::function<bool(const cv::Mat&, unsigned)>& on_frame);
+	unsigned encode_fountain(const std::string& filename, std::string output_prefix, int compression_level=6);
+	unsigned encode_fountain(const std::string& filename, const std::function<bool(const cv::Mat&, unsigned)>& on_frame, int compression_level=6);
```
sz3, line 28: Ideally/eventually these will use `cimbar::Config` to determine their defaults.

src/lib/encoder/Encoder.h
```diff
@@ -99,14 +100,34 @@ inline unsigned Encoder::encode(const std::string& filename, std::string output_
...
+inline unsigned Encoder::encode_fountain(const std::string& filename, const std::function<bool(const cv::Mat&, unsigned)>& on_frame, int compression_level)
 {
-	std::ifstream f(filename);
-	fountain_encoder_stream fes = fountain_encoder_stream::create(f, cimbar::Config::fountain_chunk_size(_eccBytes), 0); // will eventually do something clever with encode_id?
+	unsigned chunk_size = cimbar::Config::fountain_chunk_size(_eccBytes);
+
+	std::stringstream ss;
+	std::ifstream infile(filename);
+	if (compression_level <= 0)
+		ss << infile.rdbuf();
```
sz3, line 110: Because of how wirehair works, we need to load the whole file into RAM -- this was implied within `fountain_encoder_stream::create()`, but now it is explicit.

Clearly a size limit will be a good idea.

src/lib/encoder/Encoder.h
```diff
@@ -99,14 +100,34 @@ inline unsigned Encoder::encode(const std::string& filename, std::string output_
...
+	else
+	{
+		cimbar::zstd_compressor<std::stringstream> f;
+		if (!f.compress(infile))
+			return 0;
+
+		// find size of compressed zstd stream, and pad it if necessary.
+		size_t compressedSize = f.size();
+		if (compressedSize < chunk_size)
+			f.pad(chunk_size - compressedSize + 1);
```
sz3, line 120: wirehair wants the input data to need at least two chunks, and we (at least for now) would like to always use wirehair for simplicity's sake. So... pad it??

src/lib/encoder/Encoder.h
```diff
@@ -99,14 +100,34 @@ inline unsigned Encoder::encode(const std::string& filename, std::string output_
...
+		ss = std::move(f);
+	}
+
+	fountain_encoder_stream fes = fountain_encoder_stream::create(ss, chunk_size, 0); // will eventually do something clever with encode_id?
 	// With ecc = 40, we have 60 rs blocks * 115 bytes per block == 6900 bytes to work with.
 	// the fountain_chunk_size will be 690.
 	// fountain_chunks_per_frame() is currently a constant (10).
 	unsigned requiredFrames = fes.blocks_required() * 2 / cimbar::Config::fountain_chunks_per_frame();
+	if (requiredFrames == 0)
+		requiredFrames = 1; // could also do +1 on the division above?
```
sz3, line 130: The `requiredFrames` calculation in general could use some tweaking -- for `cimbar_video`, more frames is better. (and 2x or even 3 or 4 or 5x are all fine). But for outputting images to file -- right now just for testing purposes, but also presumably for the "print to paper" case -- we really shouldn't need more than a handful of extra frames.

PR #30, opened 2020-09-09
CI (travis)

sz3:
Includes a header reorder to make clang happy. (gcc tries harder to find the operator overload than clang does)


src/lib/cimb_translator/test/CimbReaderTest.cpp
```diff
@@ -20,6 +19,7 @@ namespace {
 		return s;
 	}
 }
+#include "serialize/str_join.h"
```
sz3, line 22: The `operator<<` in question is defined in the anonymous namespace right above this line.

PR #31, opened 2020-09-23
Rename `cimbar_video` to `cimbar_send`, and add GLFW rendering window (compile flag)

sz3:
Some platforms -- notably wasm -- don't have `opencv_highgui` available. The `cimbar_send` executable itself cannot be run under wasm, but is useful to manually test+verify that the GLFW-based rendering logic works

The cmake build flag is `USE_GLFW`.


CMakeLists.txt
```diff
@@ -20,13 +21,20 @@ if(NOT DEFINED OPENCV_LIBS)
...
+	add_definitions("-DLIBCIMBAR_USE_GLFW=${USE_GLFW}")
+else()
+	set(libcimbar_UI_LIB "opencv_highgui")
 endif()
 
 set (PROJECTS
 	src/exe/cimbar
-	src/exe/cimbar_scan
-	src/exe/cimbar_video
+	src/exe/cimbar_extract
```
sz3, line 36: Also renamed `cimbar_send` to `cimbar_extract`, to better reflect what you might use it for.

PR #32, opened 2020-09-24
Fix decoding bug + misc

sz3:
* fix an issue where we would happily exhaust memory with bad input data. Not ideal!
* now using `stb_image` to load our embedded symbol pngs, to limit the dependency on `opencv_imgcodecs` -- which is not available in opencv.js!
* split the fountain encoding into two steps (create encoder + encode frame), for more flexibility.


src/lib/cimb_translator/Common.cpp
```diff
@@ -46,18 +47,25 @@ namespace {
...
 
-cv::Mat load_img(string path, const string& image_dir)
+cv::Mat load_img(string path)
 {
-	if (image_dir != "")
-		return cv::imread(image_dir + "/" + path);
-
 	auto it = cimbar::bitmaps.find(path);
 	if (it == cimbar::bitmaps.end())
-		return cv::imread(path);
```
sz3, line 56: We're losing the ability to load tiles locally here, but at this point it's only needed for tests.

src/lib/cimb_translator/Common.cpp
```diff
@@ -92,9 +100,4 @@ cv::Mat getTile(unsigned symbol_bits, unsigned symbol, bool dark, unsigned num_c
 	return tile;
 }
 
-cv::Mat getTile(unsigned symbol_bits, unsigned symbol, bool dark, const string& image_dir, unsigned num_colors, unsigned color)
```
sz3, line 95: Likewise, this was only for tests, so it can go.

src/lib/encoder/Encoder.h
```diff
@@ -24,6 +24,10 @@ class Encoder
 	std::optional<cv::Mat> encode_next(STREAM& stream);
 
 	unsigned encode(const std::string& filename, std::string output_prefix);
+
+	template <typename STREAM>
+	fountain_encoder_stream::ptr create_fountain_encoder(STREAM& stream, int compression_level=6);
```
sz3, line 29: There is still some work to be done with this interface -- but instead of calling `encode_fountain`, one can now do:
```
auto fe = enc.create_fountain_encoder(instream);
auto mat = enc.encode_next(*fe);
```
... usually with the 2nd part in some kind of loop. I'll probably extend this to non-fountain encodes somehow or another.

src/lib/fountain/fountain_decoder_sink.h
```diff
@@ -78,7 +78,7 @@ class fountain_decoder_sink
 			return false;
 
 		// find or create
-		auto p = _streams.emplace(std::piecewise_construct, std::make_tuple(md.id()), std::make_tuple(md.file_size(), _chunkSize));
+		auto p = _streams.try_emplace(stream_slot(md), md.file_size(), _chunkSize);
```
sz3, line 81: `try_emplace` does what I actually want, which is to not construct the element unless it's new.

src/lib/fountain/fountain_decoder_sink.h
```diff
@@ -103,12 +103,18 @@ class fountain_decoder_sink
 		return *this;
 	}
 
+protected:
+	// streams is limited to at most 8 decoders at a time. Current, we just use the lower bits of the encode_id.
+	uint8_t stream_slot(const FountainMetadata& md) const
+	{
+		return md.encode_id() & 0x7;
```
sz3, line 110: Restrict to 8 concurrent streams...

src/lib/fountain/fountain_decoder_sink.h
```diff
@@ -103,12 +103,18 @@ class fountain_decoder_sink
...
+	}
+
 protected:
 	std::string _dataDir;
 	unsigned _chunkSize;
 
-	// streams becomes uint64_t,fds ... uint64_t will be combo of encode_id,size
-	std::map<uint64_t, fountain_decoder_stream> _streams;
-	// set just is the uint64_t
+	std::unordered_map<uint8_t, fountain_decoder_stream> _streams;
```
sz3, line 117: The old key type was `uint64_t`, which should've set off alarm bells. Oh well!

src/lib/fountain/fountain_encoder_stream.h
```diff
@@ -26,20 +30,28 @@ class fountain_encoder_stream
...
 	template <typename STREAM>
-	static fountain_encoder_stream create(STREAM& stream, unsigned buffer_size, uint8_t encode_id=0)
+	static fountain_encoder_stream::ptr create(STREAM& stream, unsigned buffer_size, uint8_t encode_id=0)
 	{
 		std::stringstream buffs;
 		if (stream)
-			 buffs << stream.rdbuf();
-		return fountain_encoder_stream(buffs.str(), buffer_size, encode_id);
+			buffs << stream.rdbuf();
+		return fountain_encoder_stream::ptr( new fountain_encoder_stream(buffs.str(), buffer_size, encode_id) );
```
sz3, line 40: Ideally this would be a constructor -- alas, the template parameter disqualifies that. The `shared_ptr` fits the use cases the best, hopefully.

src/lib/fountain/fountain_encoder_stream.h
```diff
@@ -26,20 +30,28 @@ class fountain_encoder_stream
...
 	}
 
 	bool good() const
 	{
 		return _encoder.good();
 	}
 
+	void reset()
```
sz3, line 48: Sometimes you just want to start over...

PR #34, opened 2020-10-03
Expose progress+completed data for fountain decoding

sz3:
`concurrent_fountain_decoder_sink` needs tests pretty badly at this point. Otherwise, everything seems to be working?

### DIFF
--- a/src/lib/fountain/concurrent_fountain_decoder_sink.h
+++ b/src/lib/fountain/concurrent_fountain_decoder_sink.h
@@ -11,7 +11,7 @@ class concurrent_fountain_decoder_sink
 {
 public:
 	concurrent_fountain_decoder_sink(std::string data_dir, unsigned chunk_size)
-		: _decoder(data_dir, chunk_size)
+	    : _decoder(data_dir, chunk_size)
 	{
 	}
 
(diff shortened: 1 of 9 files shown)


src/lib/fountain/concurrent_fountain_decoder_sink.h
```diff
@@ -35,14 +35,37 @@ class concurrent_fountain_decoder_sink
...
+
 	void process()
 	{
-		if (_mutex.try_lock())
+		if (_writeMutex.try_lock())
 		{
+			// maybe a 2nd mutex for retrieving _decoder.has_file()??
 			std::string buff;
 			while (_backlog.try_dequeue(buff))
 				_decoder << buff;
```
sz3, line 65: With a more flexible write interface, we would probably be able to return progress/completeness up out of the write. However, the (rough) adherence to the iostreams api prevents this. (`write()` presents the same problem)

Returning the values from the write is attractive, because it means we only need 1 mutex instead of 2 here. But I'm not going to rewrite all the streams -- yet?

src/lib/fountain/concurrent_fountain_decoder_sink.h
```diff
@@ -35,14 +35,37 @@ class concurrent_fountain_decoder_sink
...
+	{
+		std::lock_guard<std::mutex> lock(_readMutex);
+		return _progress;
+	}
+
+	void update_status()
+	{
+		// called under the writeMutex -- don't hold this for long!
+		std::lock_guard<std::mutex> lock(_readMutex);
+		_done = _decoder.get_done();
```
sz3, line 54: We could retrieve these from the streams on demand instead of caching them -- but for the moment, I'm trying to keep the read-side (the const methods) as braindead simple as possible. We want `process()` to be running unopposed.

PR #35, opened 2020-10-06
Misc fixes and improvements

sz3:
* support larger files (33MB max vs 16MB)
* add asm.js build option to go with proper wasm build
    * wasm is faster, but requires a running HTTP server to be used. Not usually annoying, but occasionally so.
* move shakycam nonsense into its own class
* add auto-incrementing `encode_id` for cimbar.js page
* performance optimizations for openGL/webGL code (used by cimbar.js)
* added cool(?) drag-drop hover effect for cimbar.js
* some mobile/tablet detection for better window sizing on platforms where you can't go fullscreen
* etc


src/lib/encoder/SimpleEncoder.h
```diff
@@ -16,18 +16,19 @@ class SimpleEncoder
 {
 public:
 	SimpleEncoder(unsigned ecc_bytes=40, unsigned bits_per_symbol=0, unsigned bits_per_color=0);
+	void set_encode_id(uint8_t encode_id); // [0-127] -- the high bit is ignored.
```
sz3, line 19: ^^^
I debated whether to steal 2 bits or 1 here, for now 1 will do.

src/lib/encoder/SimpleEncoder.h
```diff
@@ -16,18 +16,19 @@ class SimpleEncoder
...
 
 	template <typename STREAM>
 	fountain_encoder_stream::ptr create_fountain_encoder(STREAM& stream, int compression_level=6);
 
-
 protected:
 	unsigned _eccBytes;
 	unsigned _bitsPerSymbol;
 	unsigned _bitsPerColor;
+	uint8_t _encodeId = 0;
```
sz3, line 31: Ideally this would auto-increment, but it's not clear who should own the increment. The *decoder* is the part of the system that would like encode_id to be distinct (within its life cycle), but the decoder's life cycle is quite different from the encoder's!

A random encode_id (and increment from there) might be reasonable.

src/lib/fountain/FountainMetadata.h
```diff
@@ -6,6 +6,14 @@ class FountainMetadata
 public:
 	static const unsigned md_size = 4;
 
+	static void to_uint8_arr(uint8_t encode_id, unsigned size, uint8_t* arr)
+	{
+		arr[0] = (encode_id & 0x7F) | ((size >> 17) & 0x80);
```
sz3, line 11: We steal the first bit of the FountainMetadata for an extra bit of size, allowing us to address 2^25 (~33MB). This seems like a reasonable limit to me.

(notably, the fountain code implementation I'm using caps at `2^16` blocks...)

src/lib/fountain/fountain_encoder_stream.h
```diff
@@ -37,7 +38,7 @@ class fountain_encoder_stream
 		std::stringstream buffs;
 		if (stream)
 			buffs << stream.rdbuf();
-		return fountain_encoder_stream::ptr( new fountain_encoder_stream(buffs.str(), buffer_size, encode_id) );
+		return fountain_encoder_stream::ptr( new fountain_encoder_stream(buffs.str(), buffer_size, encode_id & 0x7F) );
```
sz3, line 41: This `0x7F` is extraneous, I think, but it makes me feel better.

src/lib/gui/gl_2d_display.h
```diff
@@ -15,18 +15,20 @@ class gl_2d_display
...
+	     1.0f,  1.0f, 0.0f,
+	    -1.0f,  1.0f, 0.0f
 	};
 
 public:
 	gl_2d_display()
-		: _p(create())
+	    : _p(create())
 	{
+		glGenBuffers(3, _vbo.data());
```
sz3, line 30: Creating our VBOs (and more than one of them) in advance. I don't really know what I'm doing here, but this did seem to help.

src/lib/gui/mat_to_gl.h
```diff
@@ -31,9 +29,8 @@ namespace mat_to_gl {
 			default:
 				;
 		}
-		glTexImage2D(GL_TEXTURE_2D, 0, GL_RGBA8, mat.cols, mat.rows, 0, format, GL_UNSIGNED_BYTE, mat.data);
+		glTexImage2D(GL_TEXTURE_2D, 0, GL_RGB8, mat.cols, mat.rows, 0, format, GL_UNSIGNED_BYTE, mat.data);
```
sz3, line 32: `glTexImage2D` seems finicky (at least in webGL, where we care), so we're focusing on the 24-bit RGB case that our inputs will theoretically match...

src/lib/gui/window_glfw.h
```diff
@@ -59,11 +61,8 @@ class window_glfw : public window_interface<window_glfw>
 
 		if (_display)
 		{
-			cv::Mat rgb;
-			cv::cvtColor(img, rgb, cv::COLOR_BGR2RGBA);
-			GLuint texture = cimbar::mat_to_gl::create_gl_texture(rgb);
-			_display->draw(texture);
-			glDeleteTextures(1, &texture);
+			cimbar::mat_to_gl::load_gl_texture(_texid, img);
```
sz3, line 64: no `cvtColor` -- we're hoping for RGB by default. There will be a followup PR...

Also: now calling `glGenTextures` in the constructor, and reusing it.

src/wasm/cimbar_js/CMakeLists.txt
```diff
@@ -28,13 +28,43 @@ target_link_libraries(cimbar_js
...
+	set (LINK_WASM_OPT "-s WASM=1")
+endif()
 
-install(
-	TARGETS cimbar_js
-	DESTINATION ${libcimbar_SOURCE_DIR}/web
+# extra link flags for emscripten
+set (LINK_WASM_LIST
+	--bind -Os
+	${LINK_WASM_OPT}
```
sz3, line 43: Took some minor cmake jank to get this to work, but I think this is much more readable than it was...

src/wasm/cimbar_js/CMakeLists.txt
```diff
@@ -28,13 +28,43 @@ target_link_libraries(cimbar_js
...
+string(REPLACE ";" " " LINK_WASM_FLAGS "${LINK_WASM_LIST}")
+
+if(DEFINED USE_WASM)
+	set_target_properties( cimbar_js PROPERTIES LINK_FLAGS "${LINK_WASM_FLAGS}" )
+
+	install(
+		TARGETS cimbar_js
+		DESTINATION ${libcimbar_SOURCE_DIR}/web
+	)
+	if(NOT USE_WASM EQUAL "2")
```
sz3, line 61: Apparently `EXISTS` doesn't play nice with the `$<TARGET_FILE_DIR:...` syntax?

src/wasm/cimbar_js/wasm.cpp
```diff
@@ -33,24 +35,27 @@ int initialize_GL(int width, int height)
...
 	std::optional<cv::Mat> img = enc.encode_next(*_fes);
 	if (!img)
 	{
 		std::cerr << "no image :(" << std::endl;
-		return;
+		return 0;
 	}
 
 	_window->show(*img, 0);
+	return ++_renders;
```
sz3, line 58: Currently tracking the count here, though a return 1/0 might be better.

web/main.html
```diff
@@ -35,6 +36,7 @@
...
...rgba(17, 17, 17, 0.02) 181px,rgba(179, 179, 179, 0.02) 181px, rgba(179, 179, 179, 0.02) 220px),repeating-linear-gradient(135deg, rgba(58, 58, 58, 0.02) 0px, rgba(58, 58, 58, 0.02) 41px,rgba(198, 198, 198, 0.02) 41px, rgba(198, 198, 198, 0.02) 60px,rgba(176, 176, 176, 0.02) 60px, rgba(176, 176, 176, 0.02) 99px,rgba(173, 173, 173, 0.02) 99px, rgba(173, 173, 173, 0.02) 146px,rgba(164, 164, 164, 0.02) 146px, rgba(164, 164, 164, 0.02) 167px,rgba(179, 179, 179, 0.02) 167px, rgba(179, 179, 179, 0.02) 205px,rgba(228, 228, 228, 0.02) 205px, rgba(228, 228, 228, 0.02) 230px,rgba(23, 23, 23, 0.02) 230px, rgba(23, 23, 23, 0.02) 241px),repeating-linear-gradient(135deg, rgba(190, 190, 190, 0.02) 0px, rgba(190, 190, 190, 0.02) 15px,rgba(74, 74, 74, 0.02) 15px, rgba(74, 74, 74, 0.02) 45px,rgba(98, 98, 98, 0.02) 45px, rgba(98, 98, 98, 0.02) 71px,rgba(43, 43, 43, 0.02) 71px, rgba(43, 43, 43, 0.02) 95px,rgba(131, 131, 131, 0.02) 95px, rgba(131, 131, 131, 0.02) 118px,rgba(21, 21, 21, 0.02) 118px, rgba(21, 21, 21, 0.02) 130px,rgba(77, 77, 77, 0.02) 130px, rgba(77, 77, 77, 0.02) 167px,rgba(231, 231, 231, 0.02) 167px, rgba(231, 231, 231, 0.02) 189px),linear-gradient(180deg, rgb(251, 251, 251),rgb(0,0,0));
```
sz3, line 39: cool gradient :open_mouth:

web/main.html
```diff
@@ -57,9 +59,24 @@
 	<canvas id="canvas" class="canvas"></canvas>
 </div>
 <script type="text/javascript">
+	var mobileAndTabletCheck = function() {
```
sz3, line 62: cool unreadable regex with a bunch of browsers in it :open_mouth:

web/main.html
```diff
@@ -57,9 +59,24 @@
...
+	};
+
+	// reset zoom
+	document.body.style.zoom = (window.innerWidth / window.outerWidth);
+
 	const canvas = document.getElementById('canvas');
-	canvas.width = window.outerWidth;
-	canvas.height = window.outerHeight;
+	if (mobileAndTabletCheck()) {
+		canvas.width = window.innerWidth;
```
sz3, line 73: On mobile (and tablets I guess?), use the inner sizes instead of outer sizes, because the user's already (probably) as full-screen as they can go.

web/main.js
```diff
@@ -1,6 +1,10 @@
 var Main = function() {
 
-var _interval = 100;
+var _interval = 66;
```
sz3, line 3: ~15 fps

PR #37, opened 2020-10-15
Bump "drift" limits to improve decoding of distorted images

sz3:
This goes back to #8 -- limiting our `drift` value to `[-2, 2]` turns out to be overly restrictive with better tracking of local drift values.

7 is -- as with a lot of numbers in the implementation -- probably overfit to my sample data, but also good enough for now.


src/wasm/cimbar_js/wasm.cpp
```diff
@@ -40,8 +40,11 @@ int render()
 	if (!_window or !_fes)
 		return 0;
 
-	// this should probably be x2 (at least), but for now...
-	if (_fes->block_count() > _fes->blocks_required()*2)
+	// we generate 2x the amount of required blocks -- unless everything fits in a single frame.
+	unsigned required = _fes->blocks_required();
+	if (required > cimbar::Config::fountain_chunks_per_frame())
```
sz3, line 45: also... this!

PR #40, opened 2020-10-30
Switch scanner to use OTSU by default

sz3:
* it's not a full-fledged adaptive threshold algorithm, but it seems to do pretty well. Importantly, it's a decent performance boost with throughput implications (spoilers: 93 kb/s)
   * keeping around the adaptiveThreshold logic for now -- I can imagine cases where it will be useful (paper mode?)
* fix bug in ScanState
* refactor the Scanner preprocessing a bit

### DIFF
--- a/src/lib/extractor/ScanState.h
+++ b/src/lib/extractor/ScanState.h
@@ -64,7 +64,7 @@ protected:
 			if (_tally[i] == 0)
 				return NOOP;
 
-		int center = _tally[3];
+		float center = _tally[3];
 		for (int i = 1; i <= 5; ++i)
 		{
 			if (i == 3)
--- a/src/lib/extractor/Scanner.h
+++ b/src/lib/extractor/Scanner.h
@@ -19,11 +19,14 @@ public: // public inline methods
 	template <typename MAT>
 	Scanner(const MAT& img, bool dark=true, int skip=0);
 
-	template <typename MAT>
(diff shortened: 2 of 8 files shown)


src/lib/extractor/ScanState.h
```diff
@@ -64,7 +64,7 @@ class ScanState
 			if (_tally[i] == 0)
 				return NOOP;
 
-		int center = _tally[3];
+		float center = _tally[3];
```
sz3, line 67: Integer division :disappointed:

src/lib/extractor/Scanner.h
```diff
@@ -92,40 +98,55 @@ inline unsigned Scanner::nextPowerOfTwoPlusOne(unsigned v)
...
 template <typename MAT>
-inline Scanner::Scanner(const MAT& img, bool dark, int skip)
+inline Scanner::Scanner(const MAT& img, bool fast, bool dark, int skip)
     : _dark(dark)
     , _skip(skip? skip : std::min(img.rows, img.cols) / 60)
     , _mergeCutoff(img.cols / 30)
     , _anchorSize(30)
 {
-	_img = preprocess_image(img);
+	_img = preprocess_image(img, fast);
```
sz3, line 149: The final form of this flag is probably an enum with options (`PREPROCESS_NONE, FAST, ADAPTIVE, ...`)

PR #41, opened 2020-11-02
Configurable color bits

sz3:
Still a bit "beta", but it's an improvement.


src/exe/cimbar_send/send.cpp
```diff
@@ -87,7 +90,7 @@ int main(int argc, char** argv)
 		return true;
 	};
 
-	Encoder en(ecc);
+	Encoder en(ecc, cimbar::Config::symbol_bits(), colorBits);
```
sz3, line 93: Another option is to make cimbar::Config ... run-time configurable. I might do that.

PR #52, opened 2021-02-04
Decouple the symbol and color decode to enable better caching

sz3:
`CimbDecoder.decode_color()` iterates over image pixels, which I'd noted (https://github.com/sz3/libcimbar/pull/12) can blow up the caches. One of my theories was that it might be beneficial to attempt a 2-pass approach: do the symbol decodes first, then the color decodes.

This is beneficial not just because we're doing all the color decodes at once, but because the symbol decoding occurs in a "random"-ish order (chasing the highest-confidence cells, wherever that leads). Doing the color decodes after symbol decoding is done allows us to choose what order to decode color cells in.

In practice, it seems to make for a ~10% speedup on the symbol+color decode loop.

### DIFF
--- a/src/lib/bit_file/test/bitbufferTest.cpp
+++ b/src/lib/bit_file/test/bitbufferTest.cpp
@@ -111,3 +111,36 @@ TEST_CASE( "bitbufferTest/testWriter.8", "[unit]" )
 	assertEquals( 225, bb.read(16, 8) );
 	assertEquals( 4, bb.read(24, 8) );
 }
+
+TEST_CASE( "bitbufferTest/testOverwrite", "[unit]" )
+{
+	bitbuffer bb;
+
+	bb.write(13,8064,6);
+	bb.write(7,8070,6);
+	bb.write(1,8076,6);
+	bb.write(5,8082,6);
+	bb.write(13,8088,6);
+	bb.write(7,8094,6);
+	bb.write(12,8100,6);
+	bb.write(14,8106,6);
+	bb.write(13,8112,6);
+	bb.write(7,8118,6);
+
+	bb.write(1,8064,2);
+	bb.write(0,8070,2);
+	bb.write(0,8076,2);
+	bb.write(3,8082,2);
+	bb.write(1,8088,2);
+	bb.write(0,8094,2);
+	bb.write(0,8100,2);
+	bb.write(2,8106,2);
+	bb.write(1,8112,2);
+	bb.write(0,8118,2);
+
+	assertEquals( 'u', bb.buffer()[0x3f2] );
+	assertEquals( 't', bb.buffer()[0x3f3] );
+	assertEquals( 's', (unsigned)bb.buffer()[0x3f4] );
+	assertEquals( '.', bb.buffer()[0x3f5] );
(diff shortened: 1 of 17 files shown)


src/lib/bit_file/test/bitbufferTest.cpp
```diff
@@ -111,3 +111,36 @@ TEST_CASE( "bitbufferTest/testWriter.8", "[unit]" )
...
+	bb.write(1,8076,6);
+	bb.write(5,8082,6);
+	bb.write(13,8088,6);
+	bb.write(7,8094,6);
+	bb.write(12,8100,6);
+	bb.write(14,8106,6);
+	bb.write(13,8112,6);
+	bb.write(7,8118,6);
+
+	bb.write(1,8064,2);
```
sz3, line 130: In imitation of what the `Decoder` now does.

src/lib/encoder/Decoder.h
```diff
@@ -64,14 +64,28 @@ inline unsigned Decoder::do_decode(CimbReader& reader, STREAM& ostream)
 {
 	bitbuffer bb(_bitsPerOp * 1550);
 	std::vector<unsigned> interleaveLookup = Interleave::interleave_reverse(reader.num_reads(), _interleaveBlocks, _interleavePartitions);
+	std::array<PositionData, 12400> colorPositions; // 12400 = 1500 * 8, aka the number of cells. This should probably be determined at runtime
```
sz3, line 67: This *probably* should just be a vector, but the stack is cool. We like the stack.

src/lib/encoder/Decoder.h
```diff
@@ -64,14 +64,28 @@ inline unsigned Decoder::do_decode(CimbReader& reader, STREAM& ostream)
...
 	while (!reader.done())
 	{
-		// reader should probably be in charge of the cell index (i) calculation
+		// reader is in charge of the cell index (i) calculation
 		// we can compute the bitindex ('index') here, but only the reader will know the right cell index...
-		unsigned bits = 0;
-		unsigned i = reader.read(bits);
-		unsigned bitPos = interleaveLookup[i] * _bitsPerOp;
+		PositionData pos;
+		unsigned bits = reader.read(pos);
```
sz3, line 75: The `CimbReader.read()` call now gives us symbol bits + PositionData for later.

src/lib/encoder/Decoder.h
```diff
@@ -64,14 +64,28 @@ inline unsigned Decoder::do_decode(CimbReader& reader, STREAM& ostream)
...
 		bb.write(bits, bitPos, _bitsPerOp);
+
+		colorPositions[pos.i] = {bitPos, pos.x, pos.y};
+	}
+
+	// then decode colors.
+	// the symbol+color decode could be done as one pass, but doing it as two gives us more control of the caches
+	for (const PositionData& p : colorPositions)
+	{
+		unsigned bits = reader.read_color(p);
```
sz3, line 87: Later, we read the color. We *could* sort `colorPositions` before this second loop (and I tried a few sorts), but in practice sorting by index (what the first loop does) seems to be pretty good.

Sorting by `x` is a performance disaster, which seems like more validation of the theory.

PR #55, opened 2021-03-29
Minor tweak to the decode's preprocessing "sharpen"

sz3:
rationale:
1. we're treating RGB as BGR -- per channel operations feel suspicious
2. less data to iterate over
3. seems fine on the test datasets
4. we're trying to sharpen monochrome symbols -- conceptually, color shouldn't be incredibly important

(the reasoning for 1 and 4 is somewhat at odds. In some sense the idea reduces to, "it's simpler this way")


sz3:
(note: the sharpen isn't currently used by the android app due to the extra processing cost)

PR #56, opened 2021-04-30
Remove opencv highgui code path

sz3:
This makes GLFW a requirement for native builds, but that seems reasonable given its importance for cimbar.js. `cimbar_send` also now uses the cimbar_js code path, which is built as a static library on linux.

A side effect: frame pacing improvements (= transfer speed improvements) when using `cimbar_send`.


src/exe/cimbar_send/send.cpp
```diff
@@ -29,8 +42,6 @@ int main(int argc, char** argv)
 	    ("c,colorbits", "Color bits. [0-3]", cxxopts::value<int>()->default_value(turbo::str::str(colorBits)))
 	    ("e,ecc", "ECC level", cxxopts::value<unsigned>()->default_value(turbo::str::str(ecc)))
 	    ("f,fps", "Target FPS", cxxopts::value<unsigned>()->default_value(turbo::str::str(defaultFps)))
-	    ("r,rotatecam", "Successive images are rotated 90 degrees", cxxopts::value<bool>())
-	    ("s,shakycam", "Successive images are offset, like a shaky camera effect", cxxopts::value<bool>())
```
sz3, line 33: Losing the shake/rotate flags -- I'm ok with that. The "shakycam" behavior is now standard.

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -100,22 +101,37 @@ int encode(uint8_t* buffer, size_t size)
...
 		_colorBits = color_bits;
+		_ecc = ecc;
+		_compressionLevel = compression;
 
+		// try to refresh the stream
 		if (_window and _fes)
 		{
 			unsigned buff_size_new = cimbar::Config::fountain_chunk_size(_ecc, cimbar::Config::symbol_bits() + _colorBits);
-			if (!_fes->reset_and_resize_buffer(buff_size_new))
+			if (!_fes->restart_and_resize_buffer(buff_size_new))
```
sz3, line 127: Decided to rename this -- it's usually used as a smart ptr, and `reset()` already means something.

PR #59, opened 2022-02-19
Misc bugfixes + dependency upgrades

sz3:
1. can now build against opencv4 without messing with include paths. Also, tests will pass.
2. fixed (maybe) a CFC bug where old transfers would not get cleaned up, preventing an unrelated future transfer from being attempted
    * you can just close and re-open the app though
3. upgrades:
   * catch2 (2022-1)
   * concurrentqueue (1.0.3)
   * fmt (8.1.1)
   * intx (0.7.1)
   * picosha2 (2021-8)
   * stb_image (2021-9)
   * wirehair (2021-7)

I left zstd as is -- the upgrade will need some further investigation.


src/lib/cimb_translator/CellPositions.h
```diff
@@ -1,8 +1,9 @@
 /* This code is subject to the terms of the Mozilla Public License, v.2.0. http://mozilla.org/MPL/2.0/. */
 #pragma once
 
-#include <vector>
+#include <cstddef>
```
sz3, line 4: Make modern g++ happy(er)

src/lib/fountain/FountainMetadata.h
```diff
@@ -33,7 +33,7 @@ class FountainMetadata
 		to_uint8_arr(encode_id, size, d);
 	}
 
-	unsigned id() const
+	uint32_t id() const
```
sz3, line 36: This unsigned was doing a lot of work -- the type here should've been 32bit all along.

PR #60, opened 2022-02-25
Add scripts to build asm/wasm and portable linux + add github action to draft release on tag

sz3:
I've been using a local script to do the wasm/asm part, but it's about time it made it into the repo.

The packaging scripts are designed to run in a docker container (and the CI will run it the same way) for the sake of my general sanity.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,7 +14,7 @@ jobs:
           os: ubuntu-18.04
           env: CXX="g++-7" CC="gcc-7"
 
-        - name: "linux gcc10"
+        - name: "linux gcc9"
           os: ubuntu-latest
 
         - name: "linux clang"
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -14,6 +14,9 @@ jobs:
           os: ubuntu-18.04
           env: CXX="g++-7" CC="gcc-7"
 
+        - name: "linux gcc10"
+          os: ubuntu-latest
+
         - name: "linux clang"
           os: ubuntu-18.04
           env: CXX="clang++" CC="clang"
--- a/package-portable-linux.sh
+++ b/package-portable-linux.sh
@@ -1,17 +1,31 @@
 #!/bin/sh
-# docker run --mount type=bind,source="$(pwd)",target="/usr/src/app" -it ubuntu:18.04
+## targeting old glibc
+# docker run --mount type=bind,source="$(pwd)",target="/usr/src/app" -it ubuntu:16.04
 
 cd /usr/src/app
+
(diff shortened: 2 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -14,6 +14,9 @@ jobs:
           os: ubuntu-18.04
           env: CXX="g++-7" CC="gcc-7"
 
+        - name: "linux gcc9"
+          os: ubuntu-latest
```
sz3, line 18: Also adding ubuntu-latest, which (at the moment) seems to be gcc9.

package-portable-linux.sh
```diff
@@ -0,0 +1,31 @@
+#!/bin/sh
+## targeting old glibc
+# docker run --mount type=bind,source="$(pwd)",target="/usr/src/app" -it ubuntu:16.04
```
sz3, line 3: This isn't fully automated yet, but the `objdump -T <binfile> | grep GLIBC_ | awk '{print $(NF-1)}' | sort -V` output suggests cimbar/cimbar_send built this way will work on glibc 2.14(!) and up. 16.04 is 2.23, so that's probably the more trustworthy number -- but who knows.

That said, I did test on centos7 a bit, and it works. So that's cool.

package-portable-linux.sh
```diff
@@ -0,0 +1,31 @@
...
+
+cd /usr/src/app
+
+# https://gist.github.com/jlblancoc/99521194aba975286c80f93e47966dc5
+apt update
+apt install -y software-properties-common
+add-apt-repository -y ppa:ubuntu-toolchain-r/test
```
sz3, line 10: For g++7

package-wasm.sh
```diff
@@ -0,0 +1,23 @@
+#!/bin/sh
+#docker run --mount type=bind,source="$(pwd)",target="/usr/src/app" -it emscripten/emsdk:latest
```
sz3, line 2: I'm *slightly* nervous about using `:latest` here. We'll see.

src/lib/image_hash/average_hash.h
```diff
@@ -51,6 +51,10 @@ namespace image_hash
 				uint64_t mval = (*hax) & 0x101010101ULL;
 				const uint8_t* cv = reinterpret_cast<const uint8_t*>(&mval);
 				uint8_t val = cv[0] << 4 | cv[1] << 3 | cv[2] << 2 | cv[3] << 1 | cv[4];
+				// TODO:
+				/*if (bigEndian)
```
sz3, line 55: Everything I'm targeting right now is littleendian, but this little optimization/hack will need some extra care for other platforms...

PR #61, opened 2022-03-26
Bugfix + `cimbar_recv`, a test decoder app + ...

sz3:
* the finder pattern extract logic was, simply, broken (and frustratingly, the *test* was wrong). This would lead to decodes succeeding normally from some camera angles, and being ineffective from others.
* added `cimbar_recv` for linux (maybe other desktop OSs eventually). It is very barebones.
    * example command line: `./cimbar_recv -i /dev/video2 -o /tmp`
    * notably, there's no progress indicator yet. Configuration options are minimal.
    * it's basically a toy right now. Eventually it may be more useful.
* to help with autofocus (a bit), added a grid background to cimbar.js/.org

### DIFF
--- a/src/lib/extractor/Scanner.cpp
+++ b/src/lib/extractor/Scanner.cpp
@@ -21,7 +21,7 @@ namespace {
 	{
 		if (i < 0)
 			i = N-1;
-		else if (i > N-1)
+		else if (i >= N)
 			i = 0;
 		return i;
 	}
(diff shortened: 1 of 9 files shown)


src/lib/extractor/Scanner.cpp
```diff
@@ -115,24 +115,27 @@ bool Scanner::sort_top_to_bottom(std::vector<Anchor>& anchors)
...
 		top_right = fix_index<3>(top_left - 1);
+		bottom_left = fix_index<3>(top_left + 1);
+	}
 
 	// apply the order.
-	if (&anchors[0] != &anchors[top_left])
-		std::swap(anchors[0], anchors[top_left]);
-	if (top_left != 1 and top_right != 1)
-		std::swap(anchors[1], anchors[2]);
+	anchors = {anchors[top_left], anchors[top_right], anchors[bottom_left]};
```
sz3, line 138: Ground zero -- tried to get clever, wrote a bad test, convinced myself wrong logic was right... :rage:

PR #63, opened 2022-06-06
cli update: read input filenames from stdin

sz3:
This should enable piping inputs from other processes, for example when decoding. It's not perfect -- there's not a good way to tell when `cimbar` is finished with an input (though it shouldn't take long) -- but it's a start.

Also:
* fountain mode (`-f`) is now the default. This includes zstd compression as well, and matches what cimbar.org and CFC do.
* when decoding, the `cimbar` binary will now log completed file paths to stdout after they're written
* I added a simple python(3) test for the `cimbar` binary. These API tests may be extended to testing an so/dll once a stable interface exists.

### DIFF
--- a/src/exe/cimbar/cimbar.cpp
+++ b/src/exe/cimbar/cimbar.cpp
@@ -12,6 +12,7 @@
 
 #include "cxxopts/cxxopts.hpp"
 
+#include <filesystem>
 #include <functional>
 #include <iostream>
 #include <string>
@@ -96,7 +97,6 @@ int decode(const FileIterable& infiles, const std::function<int(cv::UMat, bool, 
 	int err = 0;
 	for (const string& inf : infiles)
 	{
-		std::cerr << "processing " << inf << std::endl;
 		bool shouldPreprocess = (preprocess == 1);
 		cv::UMat img = cv::imread(inf).getUMat(cv::ACCESS_RW);
 		cv::cvtColor(img, img, cv::COLOR_BGR2RGB);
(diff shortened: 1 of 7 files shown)


src/exe/cimbar/cimbar.cpp
```diff
@@ -73,51 +173,55 @@ int main(int argc, char** argv)
...
-	    ("h,help", "Print usage")
+		("i,in", "Encoded pngs/jpgs/etc (for decode), or file to encode", cxxopts::value<vector<string>>())
+		("o,out", "Output file prefix (encoding) or directory (decoding).", cxxopts::value<string>())
+		("c,color-bits", "Color bits. [0-3]", cxxopts::value<int>()->default_value(turbo::str::str(colorBits)))
+		("e,ecc", "ECC level", cxxopts::value<unsigned>()->default_value(turbo::str::str(ecc)))
+		("z,compression", "Compression level. 0 == no compression.", cxxopts::value<int>()->default_value(turbo::str::str(compressionLevel)))
+		("color-correct", "Toggle decoding color correction. 1 == on. 0 == off.", cxxopts::value<int>()->default_value("1"))
+		("encode", "Run the encoder!", cxxopts::value<bool>())
+		("no-deskew", "Skip the deskew step -- treat input image as already extracted.", cxxopts::value<bool>())
+		("no-fountain", "Disable fountain encode/decode. Will also disable compression.", cxxopts::value<bool>())
```
sz3, line 184: fountain is now the default. The option to disable it is `--no-fountain`, in line with `--no-deskew` above.

src/exe/cimbar/cimbar.cpp
```diff
@@ -128,24 +232,32 @@ int main(int argc, char** argv)
...
+		};
+		if (useStdin)
+			return decode(StdinLineReader(), fun, no_deskew, undistort, preprocess, color_correct);
+		else
+			return decode(infiles, fun, no_deskew, undistort, preprocess, color_correct);
+	}
 
-		// else -- default case, all bells and whistles
-		fountain_decoder_sink<cimbar::zstd_decompressor<std::ofstream>> sink(outpath, chunkSize);
+	// else, the good stuff
```
sz3, line 248: The if clauses below are in a different order, but the logic is identical.

PR #73, opened 2023-06-01
Decoding improvements, and (somewhat) more flexible config for alternate grid dimensions

sz3:
This started off as an experimental branch, so it's quite long-lived and has a number of things wrapped up in it:
* bugfix for bitmatrix decode (I *think* undefined behavior due to this was the culprit for the Android 12 crash/bug...)
* fix for the build of cimbar.js on emscripten latest (until they change it again)
* added a "cooldown" functionality to the flood fill decode algorithm. As we iterate through the cells and create decode instructions for the next cell (organized by min-error on a heap), we will also track which direction we just went. e.g. if our previous cell drifted left during its decode, we will *not* drift left for our next cell. This is a simple and (probably) safe way to prevent runaway errors during the decode.
   * this only uses the adjacent cell that added us (or scheduled us) for decoding, not the other 3 adjacent cells.
   * we don't skip the center "no drift" no-op direction -- we always *hope* this is the one we're picking.
* added more seed/start positions for the flood fill decode. We were initially doing 4 -- now we are doing 8.
* added functionality to queue more decode cells based on a successful string of adjacent tile decodes
   * the idea is that if we are "right down the middle of the plate" with our decodes on the same line (no skew adjustments, low error rate), we can do better than just queuing up the adjacent 4 cells for the tile. We can also safely queue up a few cells along the same row/column we've noticed we seem to have a good handle on.
       * this acts as a layer of redundancy for finding the right location for a tile. (rather than just depending on one of the 4 adjacent tiles to be correct, we can skip over bad tiles in some cases.
   * the threshold is <= 3 error bits. This is fairly conservative for the 8x8 decode, so it may make sense to make it configurable.
* the Config class has been split up in order to support alternate configurations. The cimbar config is still specified at compile time, but the important parameters can be swapped out by adding a new struct to the "GridConf.h" file.
* various changes/fixes have been made to accomodate smaller cell sizes. 5x5 and 6x6 (vs the current 8x8) are the ones I was experimenting with.
   * this *mostly* involves changes to the decode.
* added a work-in-progress 5x5 tileset (with 4 symbols == 2 symbol bits per tile) and configuration. Some parts of this are well vetted, others are not. More improvements need to be made to the color decode before 5x5 becomes particularly usable. That said, under the right conditions, transfer rates look pretty good...


src/exe/build_image_assets/build_image_assets.cpp
```diff
@@ -29,6 +29,12 @@ map<string, string> getFileBlobs(string dir_path)
 			blobs[file_path] = base91::encode(contents);
 		}
 
+	for (int i = 0; i < 4; ++i)
+	{
+		string file_path = fmt::format("{}/2/{:02x}.png", dir_path, i);
```
sz3, line 34: I still need to add these...

src/lib/bit_file/bitmatrix.h
```diff
@@ -17,33 +17,41 @@ class bitmatrix
...
+			);
 			writer << val;
 			p += 8;
 			size -= 8;
 		}
 
 		// remainder
-		uint8_t val = 0;
-		while (size > 0)
+		if (size > 0)
```
sz3, line 36: The bug :upside_down_face:

src/lib/cimb_translator/CellDrift.cpp
```diff
@@ -29,3 +29,15 @@ void CellDrift::updateDrift(int dx, int dy)
...
 }
+
+// static
+uint8_t CellDrift::calculate_cooldown(uint8_t previous, uint8_t idx)
+{
+	if (idx == 4)
+		return 4;
+	if (idx % 2 == 0)
+		return 0xFF;
+	if ((previous xor idx) == 6) // 1 and 7, 3 and 5
```
sz3, line 40: This can also be written as four (more) `if` statements. I'm not sure which is better. Probably the comment could be improved.

src/lib/cimb_translator/CimbReader.cpp
```diff
@@ -27,18 +27,20 @@ namespace {
...
 		cv::Mat symbols;
 		cv::cvtColor(img, symbols, cv::COLOR_RGB2GRAY);
 		if (needs_sharpen)
 		{
 			sharpenSymbolGrid(symbols, symbols);
-			blockSize = 7;
+			blockSize = 3;
 		}
-		cv::adaptiveThreshold(symbols, symbols, 255, cv::ADAPTIVE_THRESH_MEAN_C, cv::THRESH_BINARY, blockSize, 0);
+		cv::adaptiveThreshold(symbols, symbols, 255, cv::ADAPTIVE_THRESH_MEAN_C, cv::THRESH_BINARY, blockSize, -5);
```
sz3, line 41: The magic formula is grayscale algorithm (e.g. what RGB formula we use -- happens to be opencv's default) + blocksize + the constant. `(3,-5)` is my current best guess, but it'll be prone to some minor overfitting.

src/lib/cimb_translator/Config.cpp
```diff
@@ -1,94 +1,38 @@
...
 #include "Config.h"
+#include <cmath>
 
 namespace cimbar {
 
-bool Config::dark()
+unsigned Config::total_cells()
 {
-	return true;
+	return std::pow(cells_per_col(), 2) - std::pow(corner_padding(), 2) * 4;
```
sz3, line 9: Most of this has been moved to the header file. The cpp is now just the "smart" helper functions.

src/lib/cimb_translator/Config.h
```diff
@@ -1,30 +1,105 @@
...
-		unsigned corner_padding();
-		unsigned interleave_blocks();
-		unsigned interleave_partitions();
-
-		unsigned fountain_chunk_size(unsigned ecc, unsigned bitspercell=0);
-		unsigned fountain_chunks_per_frame();
-		unsigned compression_level();
-	}
+	protected:
+		using GridConf = Conf8x8;
```
sz3, line 11: Where one would swap in an alternate grid config, if they wanted to do such a thing.

src/lib/cimb_translator/Config.h
```diff
@@ -1,30 +1,105 @@
...
+		}
+
+		static constexpr unsigned interleave_partitions()
+		{
+			return 2;
+		}
+
+		static constexpr unsigned fountain_chunks_per_frame()
+		{
+			return 10;
```
sz3, line 94: I think this (and interleave_partitions) needs to be in the GridConf too, and will probably do that in a followup PR.

src/lib/cimb_translator/FloodDecodePositions.cpp
```diff
@@ -21,16 +22,23 @@ void FloodDecodePositions::reset()
...
 	_heap.push({lastElem, 0});
-	_heap.push({lastElem-smallRowLen, 0});
+	_heap.push({lastElem-(smallRowLen-1), 0});
+
+	// add more seed corners?
+	uint16_t betweenMarkerBlock = smallRowLen * _cellFinder.marker_size();
+	_heap.push({betweenMarkerBlock, 1});
+	_heap.push({betweenMarkerBlock+_cellFinder.dimensions()-1, 1});
+	_heap.push({lastElem-betweenMarkerBlock, 1});
+	_heap.push({lastElem-(betweenMarkerBlock+_cellFinder.dimensions()-1), 1});
```
sz3, line 41: Our 4 additional seed positions. We could add more, if it seemed to help.

src/lib/cimb_translator/FloodDecodePositions.cpp
```diff
@@ -47,30 +55,81 @@ FloodDecodePositions::iter FloodDecodePositions::next()
...
+}
+
+int FloodDecodePositions::update(unsigned index, const CellDrift& drift, unsigned error_distance, uint8_t cooldown)
+{
+	std::array<int,4> adj = _cellFinder.find(index);
+	update_adjacents(adj, drift, error_distance, cooldown);
+
+	auto& [_, prev_error, prev_cooldown] = _instructions[index];
+	// in the case where we have consecutive high confidence cells with no drift changes,
+	// it's safe(ish) to aggressively queue a few more cells
```
sz3, line 92: ^

src/lib/cimb_translator/FloodDecodePositions.cpp
```diff
@@ -47,30 +55,81 @@ FloodDecodePositions::iter FloodDecodePositions::next()
...
+		unsigned rr = 0;
+		unsigned ll = 1;
+		unsigned dd = 2;
+		unsigned uu = 3;
+
+		int rridx = adj[rr];
+		int llidx = adj[ll];
+		if (rridx >= 0 and llidx >= 0)
+		{
+			std::array<int,4> horizon = {-1, -1, -1, -1};
```
sz3, line 104: The reason this is array<4> is simply convenience. That's the size we expect for the directional calculation (up,down,left,right), so it's reused here.

src/lib/cimb_translator/GridConf.h
```diff
@@ -0,0 +1,31 @@
...
+	{
+		static constexpr unsigned color_bits = 2;
+		static constexpr unsigned symbol_bits = 4;
+		static constexpr unsigned ecc_bytes = 30;
+		static constexpr unsigned ecc_block_size = 155;
+		static constexpr int image_size = 1024;
+
+		static constexpr unsigned cell_size = 8;
+		static constexpr unsigned cell_offset = 8;
+		static constexpr unsigned cells_per_col = 112;
```
sz3, line 29: Our grid configurations. This is the current 8x8 config.

src/lib/cimb_translator/GridConf.h
```diff
@@ -0,0 +1,31 @@
...
+	{
+		static constexpr unsigned color_bits = 2;
+		static constexpr unsigned symbol_bits = 2;
+		static constexpr unsigned ecc_bytes = 40;
+		static constexpr unsigned ecc_block_size = 216;
+		static constexpr int image_size = 988;
+
+		static constexpr unsigned cell_size = 5;
+		static constexpr unsigned cell_offset = 9;
+		static constexpr unsigned cells_per_col = 162;
```
sz3, line 16: This is the 5x5 config I've been playing with. It works somewhat, but needs improvements to the color decoding algorithm before it's ready for prime time.

src/lib/cimb_translator/test/CimbReaderTest.cpp
```diff
@@ -73,9 +73,9 @@ TEST_CASE( "CimbReaderTest/testSample", "[unit]" )
 		++count;
 	}
 
-	string expected = "0=0 99=8 12097=9 12100=0 12196=1 12197=25 12198=33 12200=5 12201=0 12295=32 "
-	        "12296=32 12297=46 12298=32 12299=34 12300=30 12301=32 12394=32 12395=57 "
-	        "12396=37 12397=38 12398=10 12399=15";
+	string expected = "0=0 99=8 600=33 710=28 711=30 821=8 822=22 823=19 934=55 11688=55 11799=25 "
```
sz3, line 76: These are regression tests. Changing the flood fill decode algorithm (new seeds, new behavior on a string of successful decodes) breaks them.

src/lib/cimbar_js/CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ set (LINK_WASM_LIST
 	-s USE_GLFW=3
 	-s FILESYSTEM=0
 	-s TOTAL_MEMORY=134217728
-	-s EXPORTED_FUNCTIONS='["_render","_next_frame","_initialize_GL","_encode","_configure"]'
+	-s EXPORTED_FUNCTIONS='["_render","_next_frame","_initialize_GL","_encode","_configure","_malloc","_free"]'
```
sz3, line 51: A change to emscripten makes this necessary. I imagine there's some situation where someone wouldn't want these exported, but I'm not sure what it is.

src/lib/encoder/Decoder.h
```diff
@@ -30,6 +30,7 @@ class Decoder
 
 protected:
 	unsigned _eccBytes;
+	unsigned _eccBlockSize;
```
sz3, line 33: I'm not sure if this actually *needs* to be a member variable. But it's somewhat moot I think.

src/lib/encoder/test/DecoderTest.cpp
```diff
@@ -56,8 +56,6 @@ TEST_CASE( "DecoderTest/testDecode.Sample", "[unit]" )
 	unsigned bytesDecoded = dec.decode(TestCimbar::getSample("6bit/4_30_f0_627_extract.jpg"), decodedFile);
 	assertEquals( 9300, bytesDecoded );
 
-	if (CV_VERSION_MAJOR == 3)
-		assertEquals( "3de927c8aa0221807a2784210160cdc17567eb587bf01233d166900aadf14bf5", get_hash(decodedFile) );
-	else // # cv4
-		assertEquals( "59ddb2516b4ff5a528aebe538a22b736a6714263a454d20e146e1ffbba36c5ae", get_hash(decodedFile) );
+	if (CV_VERSION_MAJOR == 4)
+		assertEquals( "0f74a76cb1f59df7a42449a3527d464d913d12a03bffa51d6f53828724c3feb1", get_hash(decodedFile) );
```
sz3, line 60: Another intentional regression test breakage.

src/lib/extractor/Deskewer.h
```diff
@@ -11,8 +11,8 @@
 class Deskewer
 {
 public:
-	Deskewer(unsigned total_size=1024, unsigned anchor_size=30);
-	int total_size() const;
+	Deskewer(unsigned image_size=0, unsigned anchor_size=0);
```
sz3, line 14: These will now be defer to the config.

src/lib/image_hash/ahash_result.h
```diff
@@ -67,38 +70,38 @@ class ahash_result
...
+			be.extract_tuple( be.pattern(3) ),
+			be.extract_tuple( be.pattern(4) ),
+			be.extract_tuple( be.pattern(5) ),
 			// bottom row
-			be.extract(20, 30, 40, 50, 60, 70, 80, 90),
-			be.extract(21, 31, 41, 51, 61, 71, 81, 91),
-			be.extract(22, 32, 42, 52, 62, 72, 82, 92)
+			be.extract_tuple( be.pattern(6) ),
+			be.extract_tuple( be.pattern(7) ),
+			be.extract_tuple( be.pattern(8) )
```
sz3, line 86: These were already pretty magical -- to deal with configurable cell dimensions, they are now sadly even more magical.

src/lib/image_hash/ahash_result.h
```diff
@@ -67,38 +70,38 @@ class ahash_result
...
-			be.extract(1, 11, 21, 31, 41, 51, 61, 71),
+			be.extract_tuple( be.pattern(1) ),
 			0,
 			// middle row
-			be.extract(10, 20, 30, 40, 50, 60, 70, 80),
-			be.extract(11, 21, 31, 41, 51, 61, 71, 81),
-			be.extract(12, 22, 32, 42, 52, 62, 72, 82),
+			be.extract_tuple( be.pattern(3) ),
+			be.extract_tuple( be.pattern(4) ),
+			be.extract_tuple( be.pattern(5) ),
```
sz3, line 101: This is the one we actually use -- ignoring the corners. These represent the 64 distinct bits we need for each of the 5 compares (popcnts) we do for each potential symbol for each cell decode. For 8x8 (with 16 possibilities), that's 16*5 = 80 compares.

src/lib/image_hash/average_hash.h
```diff
@@ -37,72 +37,39 @@ namespace image_hash
 		return res;
 	}
 
-	inline ahash_result special_case_fuzzy_ahash(const cv::Mat& gray, unsigned mode)
+	template <unsigned CELLSIZE>
+	inline ahash_result<CELLSIZE> fuzzy_ahash(const cv::Mat& img, uchar threshold=0, unsigned mode=ahash_result<CELLSIZE>::ALL)
```
sz3, line 41: Removed some dead code.

This, sadly, now takes a template param. Could you tell?

src/lib/image_hash/bit_extractor.h
```diff
@@ -2,13 +2,33 @@
...
+	{
+		if constexpr (iterations == 1)
+			return std::make_tuple(offset);
+		else
+			return std::tuple_cat(std::make_tuple(offset), get_offsets<iterations-1>(offset+READLEN+2));
+	}
+
+	static constexpr auto pattern(unsigned id)
+	{
+		return get_offsets(id%3 + (id/3)*(READLEN+2));
```
sz3, line 26: Some indecipherable magic. I should add a comment. The tests may help in understanding what's going on here.

The magic number `3` comes from our 2 redundant rows/cols (e.g. left,center,right) -- we convert our numbering scheme in `extract_fast()` into a real initial bitposition that we'll use to extract our 64 bits (8x8) from the 100 bits (10x10) we have.

src/lib/image_hash/test/bitExtractorTest.cpp
```diff
@@ -56,3 +90,94 @@ TEST_CASE( "bitExtractorTest/testLargerValue.2", "[unit]" )
...
+	assertEquals( "0 7 14 21 28", tuple_to_str(be.get_offsets(0)) );
+
+	assertEquals( "1 8 15 22 29", tuple_to_str(be.pattern(1)) );
+	assertEquals( "1 8 15 22 29", tuple_to_str(be.get_offsets(1)) );
+
+	assertEquals( "2 9 16 23 30", tuple_to_str(be.pattern(2)) );
+	assertEquals( "2 9 16 23 30", tuple_to_str(be.get_offsets(2)) );
+
+	assertEquals( "7 14 21 28 35", tuple_to_str(be.pattern(3)) );
+	assertEquals( "7 14 21 28 35", tuple_to_str(be.get_offsets(7)) );
```
sz3, line 133: Here we assert that the magic incantation acts as expected.

src/lib/bit_file/bitmatrix.h
```diff
@@ -17,33 +17,41 @@ class bitmatrix
...
 		{
-			const uint64_t* hax = reinterpret_cast<const uint64_t*>(p);
-			uint64_t mval = (*hax) & 0x101010101010101ULL;
+			// we're turning 1 uint64_t into 8 uint8_ts
+			uint64_t mval;
+			memcpy(&mval, p, sizeof mval);
+			mval = mval & 0x101010101010101ULL;
 			const uint8_t* cv = reinterpret_cast<const uint8_t*>(&mval);
-			uint8_t val = cv[0] << 7 | cv[1] << 6 | cv[2] << 5 | cv[3] << 4 | cv[4] << 3 | cv[5] << 2 | cv[6] << 1 | cv[7];
+			// TODO: what about endianness???
```
sz3, line 25: :thinking:

src/lib/cimb_translator/CimbReader.cpp
```diff
@@ -89,19 +91,19 @@ namespace {
...
 
 CimbReader::CimbReader(const cv::Mat& img, CimbDecoder& decoder, bool needs_sharpen, bool color_correction)
-    : _image(img)
-    , _cellSize(Config::cell_size() + 2)
-    , _positions(Config::cell_spacing(), Config::num_cells(), Config::cell_size(), Config::corner_padding())
-    , _decoder(decoder)
-    , _good(_image.cols >= Config::image_size() and _image.rows >= Config::image_size())
+	: _image(img)
+	, _cellSize(Config::cell_size() + 2)
+	, _positions(Config::cell_spacing(), Config::cells_per_col(), Config::cell_offset(), Config::corner_padding())
```
sz3, line 96: `num_cells()` was a misleading name. `cells_per_col()` (and row) is what it is.

PR #75, opened 2023-06-10
Fixes for cimbar.js, cimbar_send

sz3:
Including pinning the emscripten emsdk image...

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -23,7 +23,7 @@ jobs:
       - name: Run the build process with Docker
         uses: addnab/docker-run-action@v3
         with:
-          image: emscripten/emsdk:latest
+          image: emscripten/emsdk:3.1.39
           options: -v ${{ github.workspace }}:/usr/src/app
           shell: bash
           run: |
--- a/src/lib/cimbar_js/cimbar_js.cpp
+++ b/src/lib/cimbar_js/cimbar_js.cpp
@@ -1,6 +1,7 @@
 /* This code is subject to the terms of the Mozilla Public License, v.2.0. http://mozilla.org/MPL/2.0/. */
 #include "cimbar_js.h"
 
+#include "cimb_translator/Config.h"
 #include "encoder/SimpleEncoder.h"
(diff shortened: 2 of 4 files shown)


.github/workflows/release.yml
```diff
@@ -23,7 +23,7 @@ jobs:
       - name: Run the build process with Docker
         uses: addnab/docker-run-action@v3
         with:
-          image: emscripten/emsdk:latest
+          image: emscripten/emsdk:3.1.39
```
sz3, line 26: Staying on `latest` isn't just tempting fate: it's already broken in the past month.

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -30,6 +31,11 @@ int initialize_GL(int width, int height)
 	if (_window)
 		return 1;
 
+	// must be divisible by 4???
+	if (width % 4 != 0)
+		width += (4 - width % 4);
+	if (height % 4 != 0)
+		height += (4 - height % 4);
```
sz3, line 38: I'll admit I don't understand this at all. Some opengl nonsense, I think.

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -105,11 +111,11 @@ int configure(unsigned color_bits, unsigned ecc, int compression)
 {
 	// defaults
 	if (color_bits > 3)
-		color_bits = 2;
+		color_bits = cimbar::Config::color_bits();
 	if (ecc < 0 or ecc >= 150)
-		ecc = 30;
+		ecc = cimbar::Config::ecc_bytes();
```
sz3, line 116: Honor config defaults.

web/main.js
```diff
@@ -157,7 +157,7 @@ return {
 
   setColorBits : function(color_bits)
   {
-    Module._configure(color_bits, 30, 6);
+    Module._configure(color_bits, 255, 255);
```
sz3, line 160: *Use* config defaults.

PR #78, opened 2023-09-28
Update dependencies, Fall 2023 edition

sz3:
zstd (finally), wirehair, concurrent_queue, fmt, intx. Also includes a change to the default zstd compression level. `16` is the highest before real performance drawbacks start to appear. We don't care *too much* about those performance drawbacks (compression will happen once at the start of an encode), but it also doesn't make sense to crank it to the max and introduce weird startup lag for minimal benefit.

OTOH, any compression we can do will speed up overall transfers, so 16 > 6. I think.


src/lib/encoder/test/EncoderTest.cpp
```diff
@@ -70,7 +70,7 @@ TEST_CASE( "EncoderTest/testFountain.Compress", "[unit]" )
 	Encoder enc(30, 4, 2);
 	assertEquals( 1, enc.encode_fountain(inputFile, outPrefix) );
 
-	uint64_t hash = 0x4fd34f01ee80a28d;
+	uint64_t hash = 0x664598a460acad14;
```
sz3, line 73: Of course, various regression tests broke. :upside_down_face:

PR #81, opened 2023-11-30
Add progressive web app config, and minor bugfix for symbol decoding

sz3:
In trying to improve the color decode, I've gathered a larger dataset of real-world sample images -- and it turns out that the decoder change introduced [here](https://github.com/sz3/libcimbar/pull/73/files#diff-a45783828262edbaa3574693a4a79b73aca25184e8c75362303b6d494230359dR36) *was* overfit. Incremental learning...

The other change here is to add a `pwa.json` file to enable installing cimbar.org as a webapp. There are some drawbacks (I've seen some brutal performance in airplane mode on android that simply I don't understand), but in general it seems like a reasonable thing to have, even if it's imperfect.


src/lib/cimb_translator/CimbReader.cpp
```diff
@@ -27,13 +27,16 @@ namespace {
...
 		cv::Mat symbols;
 		cv::cvtColor(img, symbols, cv::COLOR_RGB2GRAY);
 		if (needs_sharpen)
-			sharpenSymbolGrid(symbols, symbols); // we used to change blockSize for this case -- may one day be a useful trick again?
-		cv::adaptiveThreshold(symbols, symbols, 255, cv::ADAPTIVE_THRESH_MEAN_C, cv::THRESH_BINARY, blockSize, -5);
+		{
+			blockSize = 7;
+			sharpenSymbolGrid(symbols, symbols);
+		}
+		cv::adaptiveThreshold(symbols, symbols, 255, cv::ADAPTIVE_THRESH_MEAN_C, cv::THRESH_BINARY, blockSize, 0);
```
sz3, line 39: It's not clear the best way to calculate it (do you include good and bad images? borderline ones? etc etc), but the most sensible way I have of evaluating the error rate introduced by that `-5` is to check how many of the sample images fit into certain error cutoffs.

Before and after, using blockSize=3 and different constant values (top=-5, bottom=0):

| <50  | < 100 | < 200 | < 400 | < 500 | <700 | < 1000 | total |
| ---- | ----- | ----- | ----- | ----- | ---- | ------ | ----- |
| 28   | 62    | 98    | 149   | 168   | 190  | 223    |  371  |
| 107  | 141   | 177   | 208   | 226   | 251  | 266    |  371  |

Higher is better. Though for very low error rates the difference is mostly irrelevant (the ECC will correct a 1 error or 60 with similar ease), on the whole this means that the previous change was a significant regression under certain lighting conditions -- the 371 image dataset includes various lighting scenarios -- and fine under others.

Issue #84, opened 2023-12-31
Building Static/Portable


Anonymous3-a:
If I run `sudo ./package-portable-linux.sh`, it gives me this error:
<details><summary>Expand Block [Click]</summary><pre>
./package-portable-linux.sh: line 5: cd: /usr/src/app: No such file or directory
Hit:1 cdrom://Trisquel 11.0 _aramo_ - Release amd64 (20230315) aramo InRelease
Hit:2 http://archive.trisquel.org/trisquel aramo InRelease
Hit:3 http://archive.trisquel.org/trisquel aramo-updates InRelease
Hit:4 http://archive.trisquel.org/trisquel aramo-security InRelease
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
325 packages can be upgraded. Run 'apt list --upgradable' to see them.
Reading package lists... Done
Building dependency tree... Done
Reading state information... Done
The following additional packages will be installed:
  python3-software-properties software-properties-qt
The following packages will be upgraded:
  python3-software-properties software-properties-common software-properties-qt
3 upgraded, 0 newly installed, 0 to remove and 322 not upgraded.
Need to get 119 kB of archives.
After this operation, 0 B of additional disk space will be used.
Get:1 http://archive.trisquel.org/trisquel aramo-updates/main amd64 software-properties-common all 0.99.22.8+11.0trisquel13 [64.2 kB]
Get:2 http://archive.trisquel.org/trisquel aramo-updates/main amd64 software-properties-qt all 0.99.22.8+11.0trisquel13 [25.7 kB]
Get:3 http://archive.trisquel.org/trisquel aramo-updates/main amd64 python3-software-properties all 0.99.22.8+11.0trisquel13 [29.3 kB]
Fetched 119 kB in 0s (288 kB/s)                       
(Reading database ... 228949 files and directories currently installed.)
Preparing to unpack .../software-properties-common_0.99.22.8+11.0trisquel13_all.deb ...
Unpacking software-properties-common (0.99.22.8+11.0trisquel13) over (0.99.22.6+11.0trisquel12) ...
Preparing to unpack .../software-properties-qt_0.99.22.8+11.0trisquel13_all.deb ...
Unpacking software-properties-qt (0.99.22.8+11.0trisquel13) over (0.99.22.6+11.0trisquel12) ...
Preparing to unpack .../python3-software-properties_0.99.22.8+11.0trisquel13_all.deb ...
[...]

PR #85, opened 2024-01-01
Pin "portable" linux script to use an old (known good) cmake

sz3:
Found this while refreshing my memory about #84

The other dependencies were/are de-facto pinned (since ubuntu 16.04 has passed end of support), but this one wasn't.

### DIFF
--- a/package-portable-linux.sh
+++ b/package-portable-linux.sh
@@ -14,7 +14,7 @@ apt install -y pkgconf g++-7 python-pip
 apt install -y libgles2-mesa-dev libglfw3-dev
 
 # cmake (via pip)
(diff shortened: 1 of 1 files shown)


package-portable-linux.sh
```diff
@@ -14,7 +14,7 @@ apt install -y pkgconf g++-7 python-pip
 apt install -y libgles2-mesa-dev libglfw3-dev
 
 # cmake (via pip)
-python -m pip install cmake
+python -m pip install cmake==3.21.4
```
sz3, line 17: pypi: https://pypi.org/project/cmake/3.21.4/

PR #86, opened 2024-01-09
Update index.html

Headedbranch225:
Make the code show above the background, currently I only see the background but this fixes it


sz3:
Thanks for reporting this. I have a couple questions:

1. which browser are you seeing this in?
2. is it the whole box in the middle (blue/green pattern + text + barcode) that's hidden, or just the barcode?

sz3:
Setting the z-index on the parent is mostly fine, but it does hide the intro text. My attempt is here, but I'm not sure if it resolves the problem you're seeing or not: https://github.com/sz3/libcimbar/commit/4997cac6ef0313b2874918498d7c5c28f72dac2e#diff-8f62b6ced28d3396b501d2e89a2e7cb761d16cd7dc977aebece03d4a5da5c24eR38-R41

Headedbranch225:
I am running firefox and it obly shows the background without the change being made

sz3:
That's interesting. I'm on FF 121 on linux and have a few different FF android versions I usually test with -- I don't see the issue...

However, I'll try to put something together based on your fix. (unfortunately getting the text back is going to be slightly involved...)

sz3:
Ok, trying again:
https://github.com/sz3/libcimbar/commit/8f9caa18bc8b7787831123e96b565c95a02a7303

[index.html](https://github.com/sz3/libcimbar/raw/8f9caa18bc8b7787831123e96b565c95a02a7303/web/index.html)

I'm *hoping* that will do the right thing now that there's no negative z-index. It seems to work on chrome/safari/FF on linux and android, *but* apparently some firefoxes are special. :slightly_smiling_face:

sz3:
I made https://github.com/sz3/libcimbar/pull/87, which supercedes this PR and *hopefully* resolves the issue. Let me know if it doesn't...

PR #87, opened 2024-01-21
css bugfix + package-portable-linux script tweak

sz3:
#86 noted an issue with the negative z-index on Firefox -- perhaps due to some extension. I don't know if a fix is strictly necessary (*afaik* negative z-index is ok), but in looking at the way the elements are layered I think a bit of reorganization is reasonable.

Also:
* android firefox width fix(?)
* a minor change to the package-portable-linux script inspired by #84


web/main.js
```diff
@@ -49,8 +49,7 @@ return {
   {
     // reset zoom
     var canvas = document.getElementById('canvas');
-    canvas.style.zoom = (window.innerWidth / window.outerWidth);
-    var width = window.outerWidth;
+    var width = window.innerWidth;
```
sz3, line 52: In the process of testing I found a weird interaction with `window.outerWidth` seemingly being >100% on android firefox? So we'll just use `innerWidth`...

The goal is to use 100% of the visible viewport, erring on the side of <100% if there's any ambiguity. Note that we're still using `outerHeight` (not innerHeight) below, so there is room for more rigor here.

Headedbranch225:
Is it live now or do i have to dl the html and run it?

sz3:
> Is it live now or do i have to dl the html and run it?

Not live yet, give it an hour...

sz3:
Ok, I put it "live" here: https://cimbar.org/beta/

(not on the root cimbar.org yet)

sz3:
The draft of the release is here: https://github.com/sz3/libcimbar/releases/tag/untagged-b7ec5cbedd11224b3951

sz3:
\o/

Great! Thanks for testing!

sz3:
https://cimbar.org has now been updated with the new css.

Headedbranch225:
It turns out it was caused by dark reader addon

23 Jan 2024 06:24:52 sz3 ***@***.***>:

PR #90, opened 2024-02-08
Fix some longstanding (no)-fsanitize sloppiness

sz3:
* Off by one in bitbuffer that we were only hitting in tests (hopefully)
* a malloc+delete no-no (changed the `unique_ptr` deallocator to `free`)
* explicit init of a cv::Mat() in CimbDecoderTest

I'd like to enable `-fsanitize=undefined,address` by default (at least for dev builds), but we'll need to fix some library dependencies first.

### DIFF
--- a/src/lib/bit_file/bitbuffer.h
+++ b/src/lib/bit_file/bitbuffer.h
@@ -61,15 +61,15 @@ public:
 
 	bool write(unsigned data, unsigned index, int length)
 	{
-		resize(index+length-1);
+		resize(index+length);
 
 		int currentByte = index/8;
 		int currentBit = index%8;
 
 		int nextWrite = std::min(length, 8-currentBit); // write this many bits
-		while (length > 0)
+		while (length > 0 and nextWrite > 0)
 		{
-			char bits = data >> (length - nextWrite);
+			unsigned char bits = data >> (length - nextWrite);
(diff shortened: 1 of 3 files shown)


src/lib/cimb_translator/Common.cpp
```diff
@@ -57,7 +57,7 @@ cv::Mat load_img(string path)
 	vector<unsigned char> data(bytes.data(), bytes.data() + bytes.size());
 
 	int width, height, channels;
-	std::unique_ptr<uint8_t[]> imgdata(stbi_load_from_memory(data.data(), static_cast<int>(data.size()), &width, &height, &channels, STBI_rgb_alpha));
+	std::unique_ptr<uint8_t[], void (*)(void*)> imgdata(stbi_load_from_memory(data.data(), static_cast<int>(data.size()), &width, &height, &channels, STBI_rgb_alpha), ::free);
```
sz3, line 60: A neat trick I learned from the oracle that is stackoverflow

src/lib/bit_file/bitbuffer.h
```diff
@@ -61,15 +61,15 @@ class bitbuffer
 
 	bool write(unsigned data, unsigned index, int length)
 	{
-		resize(index+length-1);
+		resize(index+length);
```
sz3, line 64: :scream:

PR #91, opened 2024-02-18
Functionality to split encoding/decoding of colors and symbols

sz3:
This operates as a distinct option from the other config parameters (grid size, tile size, number of colors, ...), and like number of colors is a runtime parameter.

The motivation is to allow the symbol decode to occur independently of the color decode, allowing
(1) the decode to succeed -- at reduced capacity -- even if the color decode fails,
(2) the color decode to use the information it has from the symbol decode to generate a better color correction matrix, which will allow the color decode to succeed in a larger variety of conditions.

As an added bonus, the *color* decode can succeed while the symbol decode fails -- but in practice should not be common (in any case, it's not ideal).

The end result here is a new version of libcimbar, `0.6.0`, and a change to the format which will be called "mode B". The `0.5.x` versions ("4-color" and "8-color") will be "mode 4C" and "mode 8C" respectively. "mode 8C" support will be removed in `0.6.0`, but 4C (the more useful one) will continue to be supported until at *least* `0.7.0`, whenever that happens.

### DIFF
--- a/src/exe/cimbar/cimbar.cpp
+++ b/src/exe/cimbar/cimbar.cpp
@@ -246,7 +246,7 @@ int main(int argc, char** argv)
 	}
 
 	// else, the good stuff
-	unsigned chunkSize = cimbar::Config::fountain_chunk_size(ecc);
+	unsigned chunkSize = cimbar::Config::fountain_chunk_size(ecc, colorBits+cimbar::Config::symbol_bits());
 	if (compressionLevel <= 0)
 	{
 		fountain_decoder_sink<std::ofstream> sink(outpath, chunkSize, true);
--- a/src/exe/cimbar/cimbar.cpp
+++ b/src/exe/cimbar/cimbar.cpp
@@ -101,9 +101,11 @@ namespace {
 }
 
 template <typename FilenameIterable>
-int encode(const FilenameIterable& infiles, const std::string& outpath, int ecc, int color_bits, int compression_level, bool no_fountain)
+int encode(const FilenameIterable& infiles, const std::string& outpath, int ecc, int color_bits, int compression_level, bool legacy_mode, bool no_fountain)
 {
 	Encoder en(ecc, cimbar::Config::symbol_bits(), color_bits);
+	if (legacy_mode)
+		en.set_legacy_mode();
 	for (const string& f : infiles)
 	{
 		if (f.empty())
@@ -179,6 +181,7 @@ int main(int argc, char** argv)
 		("o,out", "Output file prefix (encoding) or directory (decoding).", cxxopts::value<string>())
 		("c,color-bits", "Color bits. [0-3]", cxxopts::value<int>()->default_value(turbo::str::str(colorBits)))
 		("e,ecc", "ECC level", cxxopts::value<unsigned>()->default_value(turbo::str::str(ecc)))
+		("m,mode", "Select a cimbar mode. B (the default) is new to 0.6.x. 4C is the 0.5.x config. [B,4C]", cxxopts::value<string>()->default_value("B"))
 		("z,compression", "Compression level. 0 == no compression.", cxxopts::value<int>()->default_value(turbo::str::str(compressionLevel)))
 		("color-correct", "Toggle decoding color correction. 2 == full (fountain mode only). 1 == simple. 0 == off.", cxxopts::value<int>()->default_value("2"))
 		("color-correction-file", "Debug -- save color correction matrix generated during fountain decode, or use it for non-fountain decodes", cxxopts::value<string>())
@@ -219,12 +222,19 @@ int main(int argc, char** argv)
 	compressionLevel = result["compression"].as<int>();
 	ecc = result["ecc"].as<unsigned>();
 
+	bool legacy_mode = false;
+	if (result.count("mode"))
+	{
+		string mode = result["mode"].as<string>();
+		legacy_mode = (mode == "4c") or (mode == "4C");
+	}
+
 	if (encodeFlag)
 	{
 		if (useStdin)
-			return encode(StdinLineReader(), outpath, ecc, colorBits, compressionLevel, no_fountain);
+			return encode(StdinLineReader(), outpath, ecc, colorBits, compressionLevel, legacy_mode, no_fountain);
 		else
-			return encode(infiles, outpath, ecc, colorBits, compressionLevel, no_fountain);
+			return encode(infiles, outpath, ecc, colorBits, compressionLevel, legacy_mode, no_fountain);
 	}
 
 	// else, decode
@@ -236,7 +246,9 @@ int main(int argc, char** argv)
 		color_correction_file = result["color-correction-file"].as<string>();
 	int preprocess = result["preprocess"].as<int>();
 
-	Decoder d(ecc, colorBits);
+	unsigned color_mode = legacy_mode? 0 : 1;
+	bool coupled = legacy_mode;
+	Decoder d(ecc, colorBits, color_mode, coupled);
 
 	if (no_fountain)
 	{
--- a/src/exe/cimbar/cimbar.cpp
+++ b/src/exe/cimbar/cimbar.cpp
@@ -269,7 +269,7 @@ int main(int argc, char** argv)
 	// else, the good stuff
 	int res = -200;
 
-	unsigned chunkSize = cimbar::Config::fountain_chunk_size(ecc, colorBits+cimbar::Config::symbol_bits());
+	unsigned chunkSize = cimbar::Config::fountain_chunk_size(ecc, colorBits+cimbar::Config::symbol_bits(), legacy_mode);
 	if (compressionLevel <= 0)
 	{
 		fountain_decoder_sink<std::ofstream> sink(outpath, chunkSize, true);
--- a/src/exe/cimbar/cimbar.cpp
+++ b/src/exe/cimbar/cimbar.cpp
@@ -117,7 +117,7 @@ int encode(const FilenameIterable& infiles, const std::string& outpath, int ecc,
 }
 
 template <typename FilenameIterable>
-int decode(const FilenameIterable& infiles, const std::function<int(cv::UMat, bool, bool)>& decode, bool no_deskew, bool undistort, int preprocess, bool color_correct)
+int decode(const FilenameIterable& infiles, const std::function<int(cv::UMat, bool, int)>& decodefun, bool no_deskew, bool undistort, int preprocess, int color_correct)
 {
 	int err = 0;
 	for (const string& inf : infiles)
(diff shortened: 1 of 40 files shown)


src/exe/cimbar/cimbar.cpp
```diff
@@ -101,9 +101,11 @@ namespace {
 }
 
 template <typename FilenameIterable>
-int encode(const FilenameIterable& infiles, const std::string& outpath, int ecc, int color_bits, int compression_level, bool no_fountain)
+int encode(const FilenameIterable& infiles, const std::string& outpath, int ecc, int color_bits, int compression_level, bool legacy_mode, bool no_fountain)
 {
 	Encoder en(ecc, cimbar::Config::symbol_bits(), color_bits);
+	if (legacy_mode)
+		en.set_legacy_mode();
```
sz3, line 108: This is currently the toggle for the old mode (interleaved symbol/color bits). It probably needs to be refactored into a config option, but for now this works.

src/exe/cimbar/cimbar.cpp
```diff
@@ -177,8 +181,10 @@ int main(int argc, char** argv)
 		("o,out", "Output file prefix (encoding) or directory (decoding).", cxxopts::value<string>())
 		("c,color-bits", "Color bits. [0-3]", cxxopts::value<int>()->default_value(turbo::str::str(colorBits)))
 		("e,ecc", "ECC level", cxxopts::value<unsigned>()->default_value(turbo::str::str(ecc)))
+		("m,mode", "Select a cimbar mode. B (the default) is new to 0.6.x. 4C is the 0.5.x config. [B,4C]", cxxopts::value<string>()->default_value("B"))
 		("z,compression", "Compression level. 0 == no compression.", cxxopts::value<int>()->default_value(turbo::str::str(compressionLevel)))
-		("color-correct", "Toggle decoding color correction. 1 == on. 0 == off.", cxxopts::value<int>()->default_value("1"))
+		("color-correct", "Toggle decoding color correction. 2 == full (fountain mode only). 1 == simple. 0 == off.", cxxopts::value<int>()->default_value("2"))
```
sz3, line 186: Leaning towards removing the old ("simple") color-correct logic. It's nice that it can be run under more conditions, but in its current form it's just not all that useful.

src/exe/cimbar/cimbar.cpp
```diff
@@ -177,8 +181,10 @@ int main(int argc, char** argv)
...
+		("color-correction-file", "Debug -- save color correction matrix generated during fountain decode, or use it for non-fountain decodes", cxxopts::value<string>())
```
sz3, line 187: This is used for doing a two-pass manual debug/verification process:
1. decode sample in fountain mode, saving ccm to file
2. use ccm file when decoding in non-fountain mode, generating a sample output file (errors and all) that can be used to validate error rates (e.g. with the [cimbar grader tool](https://github.com/sz3/cimbar/blob/master/cimbar/grader.py) )

src/exe/cimbar/cimbar.cpp
```diff
@@ -216,48 +222,70 @@ int main(int argc, char** argv)
 	compressionLevel = result["compression"].as<int>();
 	ecc = result["ecc"].as<unsigned>();
 
+	bool legacy_mode = false;
+	if (result.count("mode"))
+	{
+		string mode = result["mode"].as<string>();
+		legacy_mode = (mode == "4c") or (mode == "4C");
```
sz3, line 229: `-m 4C` to switch back to cimbar 0.5.x mode.

src/lib/cimb_translator/CimbReader.cpp
```diff
@@ -141,6 +148,114 @@ bool CimbReader::done() const
 	return !_good or _positions.done();
 }
 
+void CimbReader::init_ccm(unsigned color_bits, unsigned interleave_blocks, unsigned interleave_partitions, unsigned fountain_blocks)
+{
+	if (_colorCorrection != 2)
+		return;
+
+	// if no fountain header, we don't attempt color correction
```
sz3, line 156: This method coordinates the new color correction logic. It needs a refactor, but the comments hopefully explain the basics (cimbar-py also contains an implementation).

src/lib/cimb_translator/Common.cpp
```diff
@@ -19,27 +20,53 @@ using std::vector;
...
 	RGB getColor4(unsigned index)
 	{
-		// opencv uses BGR, but we don't have to conform to this tyranny
+		// opencv uses BGR, but we don't have to conform to its tyranny
 		static constexpr array<RGB, 4> colors = {
-		    RGB(0, 0xFF, 0xFF),
-		    RGB(0xFF, 0xFF, 0),
-		    RGB(0xFF, 0, 0xFF),
-		    RGB(0, 0xFF, 0)
+			RGB(0, 0xFF, 0),
```
sz3, line 25: Since we're changing the format anyway, I fixed the color ordering. In practice, because reedsolomon is fixing byte errors (not bit errors), this shouldn't change anything. But it does lead to less bit errors.

sz3: As a higher level note, it might make sense to fold the dark/light distinction into these color sets. Targeting a white background will require a different color set anyway.

src/lib/cimb_translator/Common.cpp
```diff
@@ -19,27 +20,53 @@ using std::vector;
...
 		static constexpr array<RGB, 8> colors = {
-		    RGB(0, 0xFF, 0xFF), // cyan
-		    RGB(0x7F, 0x7F, 0xFF),  // mid-blue
-		    RGB(0xFF, 0, 0xFF), // magenta
-		    RGB(0xFF, 65, 65), // red
-		    RGB(0xFF, 0x9F, 0),  // orange
-		    RGB(0xFF, 0xFF, 0), // yellow
-		    RGB(0xFF, 0xFF, 0xFF),
-		    RGB(0, 0xFF, 0),
+			RGB(0, 0xFF, 0xFF), // cyan
```
sz3, line 47: No 8-color configurations are supported by 0.6.0 -- but this is the correct (minimal bit errors) order of the existing colors.

src/lib/cimb_translator/Config.h
```diff
@@ -89,12 +94,12 @@ namespace cimbar
 			return 2;
 		}
 
-		static constexpr unsigned fountain_chunks_per_frame()
+		static constexpr unsigned fountain_chunks_per_frame(unsigned bitspercell, bool legacy_mode)
 		{
-			return 10;
+			return legacy_mode? 10 : bitspercell << 1;
```
sz3, line 99: fountain_chunks_per_frame needs to be a per-config option. mode B is 12, 4C and 8C are 10.

But this little workaround/hack is a sign that this Config namespace needs a rewrite to be able to better do its job. (this also will reduce the amount of distinct params that get passed into the Decoder/Encoder/etc)

src/lib/cimb_translator/Config.h
```diff
@@ -16,6 +16,11 @@ namespace cimbar
 			return true;
 		}
 
+		static constexpr unsigned color_mode()
```
sz3, line 19: My current thought is to combine this with `dark` mode.

src/lib/cimb_translator/GridConf.h
```diff
@@ -27,5 +27,8 @@ namespace cimbar
 		static constexpr unsigned cell_size = 8;
 		static constexpr unsigned cell_offset = 8;
 		static constexpr unsigned cells_per_col = 112;
+
+		static constexpr int interleave_partitions = 2;
+		static constexpr int fountain_chunks_per_frame = 10;
```
sz3, line 32: These two are unused at present.

src/lib/encoder/Decoder.h
```diff
@@ -64,10 +76,63 @@ inline Decoder::Decoder(int ecc_bytes, int color_bits, bool interleave)
 template <typename STREAM>
 inline unsigned Decoder::do_decode(CimbReader& reader, STREAM& ostream)
 {
+	if (_coupled)
+		return do_decode_coupled(reader, ostream);
```
sz3, line 80: "do it the old way"

src/lib/encoder/Decoder.h
```diff
@@ -125,3 +198,30 @@ inline unsigned Decoder::decode(std::string filename, std::string output)
...
 }
+
+inline bool Decoder::load_ccm(std::string filename)
+{
+	File f(filename);
+	std::string data = f.read_all();
+	if (data.size() < 3*3*4)
+		return false;
+
+	cv::Mat temp(3, 3, CV_32F, data.data());
```
sz3, line 209: These save/load functions are not intended for anything beyond manual debugging/testing/verification at present. (this is for the two-pass "fountain decode, then non-fountain decode, then diff to check error rates" use case).

src/lib/encoder/Encoder.h
```diff
@@ -15,8 +15,8 @@ class Encoder : public SimpleEncoder
 	using SimpleEncoder::SimpleEncoder;
 
 	unsigned encode(const std::string& filename, std::string output_prefix);
-	unsigned encode_fountain(const std::string& filename, std::string output_prefix, int compression_level=6, double redundancy=1.2, int canvas_size=0);
-	unsigned encode_fountain(const std::string& filename, const std::function<bool(const cv::Mat&, unsigned)>& on_frame, int compression_level=6, double redundancy=4.0, int canvas_size=0);
+	unsigned encode_fountain(const std::string& filename, std::string output_prefix, int compression_level=16, double redundancy=1.2, int canvas_size=0);
```
sz3, line 18: Updated default compression (what cimbar_js will use) to 16. Arguably we could go further, but this seems to be in the right place on the performance <-> size curve.

src/lib/encoder/SimpleEncoder.h
```diff
@@ -39,34 +49,98 @@ inline SimpleEncoder::SimpleEncoder(int ecc_bytes, unsigned bits_per_symbol, int
...
- *         bits1 = bb.get(_bitsPerSymbol)
- *         bits2 = bb.get(_bitsPerColor)
- *
- * */
-
 template <typename STREAM>
 inline std::optional<cv::Mat> SimpleEncoder::encode_next(STREAM& stream, int canvas_size)
 {
+	if (_coupled)
+		return encode_next_coupled(stream, canvas_size);
```
sz3, line 72: "do it the old way"

src/lib/encoder/SimpleEncoder.h
```diff
@@ -92,7 +166,7 @@ inline std::optional<cv::Mat> SimpleEncoder::encode_next(STREAM& stream, int can
 template <typename STREAM>
 inline fountain_encoder_stream::ptr SimpleEncoder::create_fountain_encoder(STREAM& stream, int compression_level)
 {
-	unsigned chunk_size = cimbar::Config::fountain_chunk_size(_eccBytes, _bitsPerColor + _bitsPerSymbol);
+	unsigned chunk_size = cimbar::Config::fountain_chunk_size(_eccBytes, _bitsPerColor + _bitsPerSymbol, (_colorMode==0 and _coupled));
```
sz3, line 169: This is gross, and returns to that "the config object needs to be more capable" idea.

src/lib/fountain/FountainMetadata.h
```diff
@@ -1,48 +1,75 @@
...
 
+	uint16_t block_id() const
+	{
+		const uint8_t* d = data()+4;
+		uint16_t res = d[1];
+		res |= ((uint16_t)d[0] << 8);
+		return res;
+	}
+
+	void increment_block_id()
```
sz3, line 68: Updated this object to track the block_id. That we `increment_*` it is an arbitrary implementation detail -- we could also `set_*` and do the computation outside this class. It's not in the critical path, so performance is not optimal but shouldn't be an issue.

src/lib/cimb_translator/CimbReader.cpp
```diff
@@ -141,6 +148,114 @@ bool CimbReader::done() const
 	return !_good or _positions.done();
 }
 
+void CimbReader::init_ccm(unsigned color_bits, unsigned interleave_blocks, unsigned interleave_partitions, unsigned fountain_blocks)
```
sz3, line 148: This method coordinates the new color correction logic. It needs a refactor, but the comments hopefully explain the basics (cimbar-py also contains an implementation).

sz3: But to simplify:
* we use the deterministic 48-bit fountain-chunk header to sample expected color values to use as an ad-hoc color "key".
* the sampled color averages for the 4 colors are then [transformed](https://en.wikipedia.org/wiki/Moore-Penrose_inverse) into the ccm

PR #92, opened 2024-02-19
Update cimbar_js UI with "mode B" toggle + docs

sz3:
A fast-follow to #91. Combined, these will be 0.6.0.

1. instead of "4-color/8-color", the toggle is now "Mode B", "Mode 4C". The default is B.
2. long-overdue docs updates!
3. make the ccm matrix a static threadlocal. This averts a longstanding ( :upside_down_face: ) race condition in CFC where multiple threads might simultaneously update their (shared) `CimbDecoder` instance with unknowable results. This is arguably a bug in CFC, but for *now* at least I'm going to experiment with using a threadlocal.


PERFORMANCE.md
```diff
@@ -4,36 +4,48 @@
 ## Numbers of note
 
 * The barcode is `1024x1024` pixels. The individual tiles are `8x8` in a `9x9` grid (there is an empty row/column of spacing on either side)
-* **7500** or 8750 bytes per cimbar image, after error correction
+* **7500** bytes per cimbar image, after error correction
 	* There are 16 possible symbols per tile, encoding 4 bits
 	* There are 4 or 8 possible colors, encoding an additional 2-3 bits per tile.
-	* These 6-7 bits per tile work out to a maximum of 9300-10850 bytes per barcode, though in practice this number is reduced by error correction.
+	* These 6 bits per tile work out to a maximum of 9300 bytes per barcode, though in practice this number is reduced by error correction.
```
sz3, line 10: Removing 8-color mode numbers.

PERFORMANCE.md
```diff
@@ -4,36 +4,48 @@
...
 	* The default ecc setting is 30/155, which is how we go from 9300 -> 7500 bytes of real data for a 4-color cimbar image.
 		* Reed Solomon is not perfect for this use case -- specifically, it corrects byte errors, and cimbar errors tend to involve 1-3 bits at a time. However, since Reed Solomon implementations are ubiquitous, it is currently in use.
 
 ## Current sustained benchmark
 
-* 4-color cimbar with ecc=30:
+* `mode B` (8x8 4-color) cimbar with ecc=30/155:
+	* 4,689,084 bytes (after compression) in 44s -> 852 kilobits/s (~106 KB/s)
```
sz3, line 17: I'm actually using the same sample file, but the compression level is different. I was encouraged to see mode B holding its own, though I do think it might be slightly more sensitive than the old mode (4C) as far as "best case" performance is concerned.

PERFORMANCE.md
```diff
@@ -4,36 +4,48 @@
...
 	* 4,717,525 bytes (after compression) in 45s -> 838 kilobits/s (~104 KB/s)
+	* the original configuration. Mostly replaced by mode B.
 
-* 8-color cimbar with ecc=30:
+* *deprecated* `mode 8C` (8x8 8-color) cimbar with ecc=30/155:
 	* 4,717,525 bytes in 40s -> 943 kilobits/s (~118 KB/s)
+	* removed in 0.6.0. 8-color has always been inconsistent, and needs future research
+
+* *beta* `mode S` (5x5 4-color) cimbar with ecc=40/216 (note: not finalized, and requires a special build)
+	* safely >1 Mbit/s
```
sz3, line 29: A teaser!

PERFORMANCE.md
```diff
@@ -4,36 +4,48 @@
...
+		* more modern cell CPUs run the decoder more quickly, but it turns out that this does not benefit performance much: the camera is usually the bottleneck.
+	* the sender is the cimbar.org wasm implementation. An equivalent command line is `./cimbar_send /path/to/file`
 		* cimbar.org uses the `shakycam` option to allow the receiver to detect/discard "in between" frames as part of the scan step. This allows it to spend more processing time decoding real data.
 	* burst rate can be higher (or lower)
-		* to this end, lower ecc settings *can* provide better burst rates
-	* 4-color cimbar is currently preferred, and will give more consistent transfer speeds.
-	* 8-color cimbar should be considered a prototype within a prototype. It is considerably more sensitive to lighting conditions and color tints.
+		* to this end, lower ecc settings *can* provide better burst rates. I've aimed for a balance of performance and reliability.
+	* cimbar `mode B` is preferred, and should be the most reliable.
+	* The older `mode 4C` *may* give more consistent transfer speeds in certain scenarios, but is mostly included for backwards-compatibility reasons.
```
sz3, line 41: .

PERFORMANCE.md
```diff
@@ -4,36 +4,48 @@
...
 
 * other notes:
-	* having better lighting in the frame often leads to better results -- this is why cimbar.org has a (mostly) white background. cfc uses android's auto-exposure, auto-focus, etc (it's a very simple app). Good ambient light -- or a white background -- can lead to more consitent quality frame capture.
+	* having better lighting in the frame often leads to better results -- this is why cimbar.org has a (mostly) white background. cfc uses android's auto-exposure, auto-focus, etc (it's a demo app). Good ambient light -- or a white background -- can lead to more consitent quality frame capture.
+		* screen brightness on the sender is good, but ambient light is better.
```
sz3, line 45: When benchmarking mode B, I was reminded of the importance of ambient lighting for optimal performance (i.e. being in a brighter room). I don't know/remember enough about optics to say whether this is a function of auto-exposure behavior, or a more fundamental physical constraint.

PERFORMANCE.md
```diff
@@ -4,36 +4,48 @@
...
 	* because of the lighting/exposure question, landscape *may* be better than portrait.
-	* cfc currently has a low resolution, so the cimbar frame should take up as much of the display as possible (trust the guide brackets)
+	* the cimbar frame should take up as much of the display as possible (trust the guide brackets)
+		* the format is designed to decode at resolutions as low as 700x700, but performance may suffer.
```
sz3, line 48: In running some experiments for the 5x5 configuration (and 8x8 as the control), I was able to shrink the encoded image incrementally down to 700x700 without seeing significant performance degradation. This was a loose goal of the initial 8x8, but I wasn't sure where the limit actually was.

README.md
```diff
@@ -5,10 +5,10 @@
 
 Behold: an experimental barcode format for air-gapped data transfer.
 
-It can sustain speeds of 943+ kilobits/s (~118 KB/s) using just a computer monitor and a smartphone camera!
+It can sustain speeds of 850 kilobits/s (~106 KB/s) using just a computer monitor and a smartphone camera!
```
sz3, line 8: Removing 8-color means we also slow down on our benchmark here. Alas.

But this should be more representative than quoting an 8-color number most people probably couldn't get working -- so I prefer it.

src/exe/cimbar/cimbar.cpp
```diff
@@ -177,6 +177,7 @@ int main(int argc, char** argv)
 	unsigned compressionLevel = cimbar::Config::compression_level();
 	unsigned ecc = cimbar::Config::ecc_bytes();
 	options.add_options()
+		("n,encode", "Run the encoder!", cxxopts::value<bool>())
```
sz3, line 180: I'm not sure about `-n`, but `-e` is already used for `ecc`. And I've seen unixtools commit worse crimes against the alphabet than this.

src/lib/cimb_translator/CimbDecoder.cpp
```diff
@@ -65,18 +65,23 @@ CimbDecoder::CimbDecoder(unsigned symbol_bits, unsigned color_bits, unsigned col
...
 }
 
-const color_correction& CimbDecoder::get_ccm() const
+// protected
+color_correction& CimbDecoder::internal_ccm() const
 {
-	// testing purposes only.
-	// this returning a thread local would be fine, iff we only use it for debugging!
+	static thread_local color_correction _ccm;
 	return _ccm;
```
sz3, line 72: The only interesting code change on this branch.

We've had a subtle(?) race condition in the CFC app ever since I introduced color correction -- this `CimbDecoder` object is shared across threads, which means that updating the ccm matrix is asking for trouble.

I could pass it around as a function param (implicitly threadlocal), but I want to experiment with static threadlocal a bit -- it might be useful for the `Config` object...

web/index.html
```diff
@@ -141,38 +141,38 @@
   margin-top: 3px;
 }
 #nav-container .icon-bar:nth-of-type(1) {
-  background-color: #00ffff;
-  box-shadow: 0px 0px 5px #00ffff, 0px 0px 2px #00ffff;
+  background-color: #00ff00;
```
sz3, line 144: All this is to change the icons when we have mode B selected vs mode 4C. It's so subtle (swapping green/cyan) that I'm not sure that it's worth it. *But* having green up top and magenta at the bottom reflects the color order we're actually using, and I couldn't resist.

PR #93, opened 2024-02-26
Switch encoder default back to 4C for now

sz3:
Follow up to #91 and #92, which introduce a new cimbar configuration ("mode B") and make it the default.

While it *will* become the default, the migration to the new mode should be easier if we do it in steps. Old apps are still expecting the old mode, after all. So, for 0.6.0 we'll keep the old 4-color config ("mode 4C") as the default for cimbar_js and cimbar_send (the encoder pieces).

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -271,7 +271,7 @@ body {
 
 <input style="display:none;" type="file" name="file_input" id="file_input" onchange="Main.fileInput(this);" />
 
-<div id="nav-container" class="mode-b">
+<div id="nav-container" class="mode-4c">
     <div class="bg" onclick="Main.blurNav();"></div>
     <button id="nav-button" tabindex="0" onclick="Main.clickNav();">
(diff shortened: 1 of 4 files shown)


web/index.html
```diff
@@ -286,7 +286,7 @@
         <li><a href="javascript:void(0)" onclick="Main.toggleFullscreen()">Fullscreen</a></li>
         <hr>
         <li><a class="mode-b" href="javascript:void(0)" onclick="Main.setMode('B')">Mode B</a></li>
-        <li><a class="mode-4c" href="javascript:void(0)" onclick="Main.setMode('4C')">Mode 4C</a></li>
+        <li><a class="mode-4c" href="javascript:void(0)" onclick="Main.setMode('4C')">4C <small>(old apps)</small></a></li>
```
sz3, line 289: Also added a bit of flavor text to the options menu.

PR #94, opened 2024-03-01
Update decoder objects for per-frame mode switching

sz3:
Motivation: per-frame bouncing between color_mode, so we can test success/failure for mode B/4C and pick the one that's working.

Details:
When running a decoder across multiple frames, we have certain state (loaded hashes for imagehash hamming distance, for example) that is longer lived -- that we don't want to run each frame. This *currently* is stored in the CimbDecoder (also the fountain_decoder_sink, though that's not relevant to this change). To allow us to switch between modes, we need the CimbDecoder to not store color_mode, but rather take it as an argument where required.

Misc:
There are other ways to accomplish this, but this change seems like the best immediate path forward. The actual technical requirement is:
(1) to run in different modes
(2) on different threads
(3) at the same time.

### DIFF
--- a/src/lib/cimb_translator/CimbDecoder.cpp
+++ b/src/lib/cimb_translator/CimbDecoder.cpp
@@ -54,11 +54,10 @@ namespace {
 	}
 }
 
-CimbDecoder::CimbDecoder(unsigned symbol_bits, unsigned color_bits, unsigned color_mode, bool dark, uchar ahashThreshold)
+CimbDecoder::CimbDecoder(unsigned symbol_bits, unsigned color_bits, bool dark, uchar ahashThreshold)
 	: _symbolBits(symbol_bits)
 	, _numSymbols(1 << symbol_bits)
 	, _numColors(1 << color_bits)
-	, _colorMode(color_mode)
 	, _dark(dark)
 	, _ahashThreshold(ahashThreshold)
 {
@@ -160,12 +159,12 @@ unsigned CimbDecoder::check_color_distance(std::tuple<uchar,uchar,uchar> a, std:
 	return color_diff(a, b);
 }
 
-std::tuple<uchar,uchar,uchar> CimbDecoder::get_color(int i) const
+std::tuple<uchar,uchar,uchar> CimbDecoder::get_color(int i, unsigned color_mode) const
 {
-	return cimbar::getColor(i, _numColors, _colorMode);
+	return cimbar::getColor(i, _numColors, color_mode);
 }
 
-unsigned CimbDecoder::get_best_color(float r, float g, float b) const
+unsigned CimbDecoder::get_best_color(float r, float g, float b, unsigned color_mode) const
 {
 	// transform color with ccm
 	if (internal_ccm().active())
@@ -188,7 +187,7 @@ unsigned CimbDecoder::get_best_color(float r, float g, float b) const
 	float best_distance = 1000000;
 	for (unsigned i = 0; i < _numColors; ++i)
 	{
-		std::tuple<uchar,uchar,uchar> candidate = get_color(i);
+		std::tuple<uchar,uchar,uchar> candidate = get_color(i, color_mode);
 		unsigned distance = check_color_distance(c, candidate);
 		if (distance < best_distance)
 		{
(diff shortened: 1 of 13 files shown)


src/lib/cimb_translator/CimbDecoder.cpp
```diff
@@ -208,12 +207,12 @@ std::tuple<uchar,uchar,uchar> CimbDecoder::avg_color(const Cell& color_cell) con
 	return center.mean_rgb();
 }
 
-unsigned CimbDecoder::decode_color(const Cell& color_cell) const
+unsigned CimbDecoder::decode_color(const Cell& color_cell, unsigned color_mode) const
```
sz3, line 210: My least favorite part of the changeset...

src/lib/cimb_translator/CimbReader.cpp
```diff
@@ -90,29 +90,30 @@ namespace {
...
-CimbReader::CimbReader(const cv::Mat& img, CimbDecoder& decoder, bool needs_sharpen, int color_correction)
+CimbReader::CimbReader(const cv::Mat& img, CimbDecoder& decoder, unsigned color_mode, bool needs_sharpen, int color_correction)
 	: _image(img)
 	, _fountainColorHeader(0U)
 	, _cellSize(Config::cell_size() + 2)
 	, _positions(Config::cell_spacing(), Config::cells_per_col(), Config::cell_offset(), Config::corner_padding())
 	, _decoder(decoder)
 	, _good(_image.cols >= Config::image_size() and _image.rows >= Config::image_size())
 	, _colorCorrection(color_correction)
+	, _colorMode(color_mode)
```
sz3, line 101: CimbReader is created per frame (the `_image(img)` is a hint of that), so storing color_mode is no problem.

src/lib/cimb_translator/test/CimbReaderTest.cpp
```diff
@@ -52,12 +52,48 @@ TEST_CASE( "CimbReaderTest/testReadOnce", "[unit]" )
...
 }
 
-TEST_CASE( "CimbReaderTest/testSample", "[unit]" )
+TEST_CASE( "CimbReaderTest/testSample.colormode0", "[unit]" )
 {
 	cv::Mat sample = TestCimbar::loadSample("6bit/4color_ecc30_fountain_0.png");
 
 	CimbDecoder decoder(4, 2);
-	CimbReader cr(sample, decoder);
+	CimbReader cr(sample, decoder, 0);
```
sz3, line 60: This is the old test case, before #91 . As long as we're supporting mode 4C, it makes sense to keep a few tests around to make sure it doesn't break.

src/lib/encoder/Decoder.h
```diff
@@ -161,27 +156,27 @@ inline unsigned Decoder::do_decode_coupled(CimbReader& reader, STREAM& ostream)
 }
 
 template <typename MAT, typename STREAM>
-inline unsigned Decoder::decode(const MAT& img, STREAM& ostream, bool should_preprocess, int color_correction)
+inline unsigned Decoder::decode(const MAT& img, STREAM& ostream, unsigned color_mode, bool should_preprocess, int color_correction)
 {
-	CimbReader reader(img, _decoder, should_preprocess, color_correction);
-	return do_decode(reader, ostream);
+	CimbReader reader(img, _decoder, color_mode, should_preprocess, color_correction);
+	return do_decode(reader, ostream, color_mode==0);
```
sz3, line 162: This `color_mode==0` check is what determines where we do the "coupled" decode -- I don't think it *really* makes sense to lock them together that way long term, so this is something that will hopefully be addressed by future changes to the way runtime configuration is done.

src/lib/encoder/test/EncoderRoundTripTest.cpp
```diff
@@ -41,11 +41,47 @@ TEST_CASE( "EncoderRoundTripTest/testFountain.Pad", "[unit]" )
...
+	assertEquals( hash, image_hash::average_hash(encodedImg) );
+
+	// decoder
+	Decoder dec(30);
+	// sink with a mismatched fountain_chunk_size
+	// importantly, the sink expects a *larger* chunk than we'll give it...
+	fountain_decoder_sink<cimbar::zstd_decompressor<std::ofstream>> fds(tempdir.path(), cimbar::Config::fountain_chunk_size(30, 6, true));
+
+	unsigned bytesDecoded = dec.decode_fountain(encodedImg, fds, 1);
+	assertEquals( 7500, bytesDecoded );
```
sz3, line 82: The key here is that the fountain decode actually *succeeds*, the aggregator (the fountain_decoder_sink) just isn't prepared for it.

sz3: ... but with further testing, it looks like this test case isn't sufficient :upside_down_face:

sz3: Error was a result of feeding bad data to wirehair -- resolved by https://github.com/sz3/libcimbar/pull/94/commits/f1e136f143ec87c8a751cdbc28d3403cd2db07d4

Issue #95, opened 2024-03-13
WebGL issue


be4zad:
My GPU is old and doesn't much support GL related things. so the website doesn't show the barcode and this error appears in the console:
```
Failed to create WebGL context: WebGL creation failed: 
* tryNativeGL (FEATURE_FAILURE_EGL_NO_CONFIG)
* Exhausted GL driver options. (FEATURE_FAILURE_WEBGL_EXHAUSTED_DRIVERS)
```

My browser is firefox and the issue may solve by:

* Enabling hardware acceleration
* Toggling webgl.out-of-process in about:config (worked for me)
* Toggling webgl.force-enabled in about:config

It's good idea to handle the error and suggest to user to try above suggestions.

sz3:
Yeah, error handling should be better. At the very *least* it'd be good to report the `WebGL creation failed` part back.

sz3:
I believe the error reporting is better in the current version (https://github.com/sz3/libcimbar/releases/tag/v0.6.1f)

PR #97, opened 2024-06-28
Misc tweaks/updates to cimbar.js encoder UI

sz3:
A bunch of small stuff:
* error message is now displayed when webGL fails init (https://github.com/sz3/libcimbar/issues/95)
* toggling the menu now temporarily pauses barcode animation for 15 frames (1 second)
    * of all the tricks I tried to get auto-focus to work better, pausing the animation is far and away the best
* on mobile, the touch event (press the screen anywhere) will also pause the feed
* new background. Very exciting and cool.
* auto-expand canvas to fill window on larger displays (was previously capped to the cimbar res of 1024x1024 +padding).
    * long overdue, probably
* make sure there's always minimum padding around the barcode

### DIFF
--- a/web/main.js
+++ b/web/main.js
@@ -1,7 +1,7 @@
 var Main = function() {
 
 var _interval = 66;
-var _pause = false;
+var _pause = 0;
 
 var _showStats = false;
 var _counter = 0;
@@ -68,17 +68,23 @@ return {
   toggleFullscreen : function()
   {
     toggleFullscreen().then(Main.resize);
-    Main.togglePause(false);
+    Main.togglePause(true);
   },
 
   togglePause : function(pause)
   {
+    // pause is a cooldown. We pause to help autofocus, but we don't want to do it forever...
     if (pause === undefined) {
-       _pause = !_pause;
+       _pause = Main.isPaused()? 0 : 32;
     }
     else {
-       _pause = pause;
+       _pause = pause? 32 : 0;
     }
+  },
+
+  isPaused : function()
+  {
+     return _pause > 0;
   },
 
   scaleCanvas : function(canvas, width, height)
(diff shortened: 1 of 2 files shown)


web/main.js
```diff
@@ -43,21 +44,45 @@ return {
...
+    }
   },
 
   resize : function()
   {
     // reset zoom
     var canvas = document.getElementById('canvas');
-    var width = window.innerWidth;
-    var height = window.outerHeight;
+    var width = window.innerWidth - 12;
```
sz3, line 62: Magic hardcoded value that (roughly) matches the css border.

web/main.js
```diff
@@ -43,21 +44,45 @@ return {
...
+    Main.togglePause(true);
+  },
+
+  togglePause : function(pause)
+  {
+    // pause is a cooldown. We pause to help autofocus, but we don't want to do it forever...
+    if (pause === undefined) {
+       pause = !Main.isPaused();
+    }
+    _pause = pause? 15 : 0;
```
sz3, line 80: The 15 is our cooldown frames.

web/main.js
```diff
@@ -144,6 +177,9 @@ return {
       _renderTime += elapsed;
       Main.setHTML( "status", elapsed + " : " + frameCount + " : " + Math.ceil(_renderTime/frameCount));
     }
+    if ( !(_counter & 31) ) {
```
sz3, line 180: A bit of a hack -- since we're doing some trickery to make the canvas seem clickable, we need to manually reposition the (absolute positioned) click element if something changes. So we have to check it on an interval.

web/main.js
```diff
@@ -194,6 +230,10 @@ window.addEventListener('keydown', function(e) {
       Main.clickNav();
       e.preventDefault();
     }
+    else if (e.key == 'Backspace' || e.keyCode == 8) {
```
sz3, line 233: You can also pause the animation with `backspace` if you have a keyboard available (i.e. on desktop).

sz3:
Added https://github.com/sz3/libcimbar/commit/75d6cfb77bf5d75df0a9498d546120019e9e522e directly to master to fix up the padding (to -10 from -12) and use innerHeight.

PR #106, opened 2024-09-22
chore: update stb_image.h

eltociear:
exitting -> exiting


sz3:
Thanks, but... this is a 3rd party dependency (from here: https://github.com/nothings/stb), and I'd prefer to keep it 100% identical to upstream (including typos).

That said, the version we have might be a bit out of date...

PR #112, opened 2024-10-21
Upgrade opencv + docker image used for cimbar.js wasm build

sz3:
* opencv to 4.10
* pin newer emscripten docker image
* minor tweak to the custom wirehair cmakelists to make sure -march is left off of the wasm build
   * might end up turning it off entirely? :thinking:

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -16,14 +16,14 @@ jobs:
 
       - name: Get openCV
         run: |
-          wget https://github.com/opencv/opencv/archive/refs/tags/4.5.5.zip
-          unzip 4.5.5.zip
-          mv opencv-4.5.5 opencv4
+          wget https://github.com/opencv/opencv/archive/refs/tags/4.10.0.zip
+          unzip 4.10.0.zip
+          mv opencv-4.10.0 opencv4
 
       - name: Run the build process with Docker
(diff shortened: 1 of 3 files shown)


src/third_party_lib/wirehair/CMakeLists.txt
```diff
@@ -54,7 +54,7 @@ set(GEN_TABLES
         gf256.h
         )
 
-if(NOT ANDROID)
+if(NOT ANDROID AND NOT USE_WASM)
     set(ARCH_NATIVE "-march=native")
```
sz3, line 58: The reason for pause I have is that low-power devices (i.e. android phones) are where we'd really benefit from the optimization -- and we're not doing it there anyway. So it might make sense to toss this altogether.

.github/workflows/release.yml
```diff
@@ -16,14 +16,14 @@ jobs:
 
       - name: Get openCV
         run: |
-          wget https://github.com/opencv/opencv/archive/refs/tags/4.5.5.zip
-          unzip 4.5.5.zip
-          mv opencv-4.5.5 opencv4
+          wget https://github.com/opencv/opencv/archive/refs/tags/4.10.0.zip
```
sz3, line 19: https://github.com/opencv/opencv/releases/tag/4.10.0

Issue #108, opened 2024-09-28
decode error


pyke8351:
Excuse me. I've run into a problem when using the decoding function. The bytes in this line of code is 0: 
# int bytes = decodefun(img, color_mode, shouldPreprocess, color_correct);
Among them, img can be loaded correctly.
color_mode: 1.
shouldPreprocess: 1
color_correct: 2
The encoding function is normal.

sz3:
What's the context here? Is the decode/encode being attempted via the `cimbar` command line tool?

For reference, that function signature is here: 
https://github.com/sz3/libcimbar/blob/75d6cfb77bf5d75df0a9498d546120019e9e522e/src/exe/cimbar/cimbar.cpp#L259-L265

... which is a decode in the basic, non-fountain format. In other words it does the reedsolomon error correction decoding but not wirehair. This is mostly useful for debugging -- if we have a single image, we can see if the reedsolomon part of the decode succeeds or not.

If it doesn't, that can mean a few things:
1. maybe the image failed to extract. Some individual cimbar images act pathological against the extract algorithm, making them borderline useless. (for reference, this is the same concern that leads to QR codes having multiple mask patterns. cimbar doesn't have the same tricks, but because it's a stream we can brute force it)
    * In the context of the encoder generating a redundant set of barcodes this is not a big deal, but for an individual image it can be an issue.
    * I've considered adding an extract sanity check to the cli encode tool. The idea would be to discard "bad" codes, since we can just ask the fountain encoder for the next in the sequence anyway.
2. maybe the parameters don't match. For example, if the encode used mode 4C and the decode is using mode B, the decode will fail. Similarly, color_mode must also be correct (the cli should handle this if the mode is correct)
3. the image could be bad, blurry, etc

pyke8351:
Thank you for your reply.I employed the command-line input, namely “./cimbar --encode -i inputfile.txt -o outputprefix” as mentioned in the example, to obtain the image. When I attempted to decode it, no result was yielded, just as I previously stated. Nevertheless, when I utilized the app for decoding, the corresponding file could be obtained. Maybe I am not yet quite familiar with the code and I should study it more earnestly.

pyke8351:
What I need to supplement and explain is that the aforesaid situation occurred when the size of my file was less than 128 bytes, and it operated normally when the file size was greater than 128 bytes.

pyke8351:
When the file is too small, for example, less than 183 bytes, it is likely to encounter the above problems.

[...]

PR #114, opened 2024-11-25
Have the (cli) encoder skip unextractable frames

sz3:
In investigating what might be causing #108, I remembered an error case I'd run into in the past but had thought/assumed/hoped was very rare: generating an "unextractable" frame -- that is, one which `cimbar_extract` will fail to extract (notwithstanding a hypothetical perfect extraction algorithm). This happens due to the extraction algorithm getting confused by data part of the code and failing to find the corner markers.

Turns out it's rare, but not rare enough. In my testing, anywhere between 1-2% (!) of generated frames are bogus in this way. This is *ok* for purposes of the main use case -- there will be another unique frame every 66ms or so. But for offline cases where we generate an image (or series of images) and check it later, it's *not* ok.

QR codes address this issue by having different "masking patterns" that can be selected. We don't have that option, but we do have a way to fix it: check the generated image after we generate, and make sure it extracts. If it doesn't, we skip that frame and try again.

Currently this is only used in the `cimbar` cli, not in the web encoder. I'll need to look at the performance implications some more to see if having the web encoder make the same check is a good idea.

### DIFF
--- a/src/lib/encoder/Encoder.h
+++ b/src/lib/encoder/Encoder.h
@@ -3,6 +3,7 @@
 
 #include "SimpleEncoder.h"
 #include "cimb_translator/Config.h"
+#include "extractor/Scanner.h"
 #include "serialize/format.h"
 
 #include <opencv2/opencv.hpp>
@@ -60,6 +61,13 @@ inline unsigned Encoder::encode_fountain(const std::string& filename, const std:
 		auto frame = encode_next(*fes, canvas_size);
 		if (!frame)
 			break;
+
+		// some % of generated frames (for the current 8x8 impl)
+		// will produce random patterns that falsely match as
+		// corner "anchors" and fail to extract. So:
+		// if frame fails the scan, skip it.
+		if (!Scanner::will_it_scan(*frame))
+			continue;
 
 		if (!on_frame(*frame, i))
 			break;
--- a/src/lib/encoder/Encoder.h
+++ b/src/lib/encoder/Encoder.h
@@ -56,6 +56,7 @@ inline unsigned Encoder::encode_fountain(const std::string& filename, const std:
 		requiredFrames = 1;
 
 	unsigned i = 0;
+	unsigned consecutiveScansFailed = 0;
 	while (i < requiredFrames)
 	{
 		auto frame = encode_next(*fes, canvas_size);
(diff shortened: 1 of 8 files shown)


src/lib/extractor/Scanner.h
```diff
@@ -141,10 +166,10 @@ inline void Scanner::preprocess_image(const MAT& img, MAT2& out, bool fast)
...
 template <typename MAT>
 inline Scanner::Scanner(const MAT& img, bool fast, bool dark, int skip)
-    : _dark(dark)
-    , _skip(skip? skip : std::min(img.rows, img.cols) / 60)
-    , _mergeCutoff(img.cols / 30)
-    , _anchorSize(30)
+	: _dark(dark)
+	, _skip(skip? skip : std::min(img.rows, img.cols) / 60)
+	, _mergeCutoff(img.cols / 30)
+	, _anchorSize(30)
```
sz3, line 172: Thinking it might be time to `clang-format` the source tree...

src/lib/extractor/Scanner.h
```diff
@@ -98,6 +102,27 @@ inline unsigned Scanner::nextPowerOfTwoPlusOne(unsigned v)
...
+inline bool Scanner::will_it_scan(const MAT& unpadded_img)
+{
+	Scanner scanner(unpadded_img);
+	std::vector<Anchor> points = scanner.scan();
+	if (points.size() < 4)
+		return false;
+
+	constexpr int limit = 50;
+	Corners corners(points);
+	if (corners.top_left().x() > limit or corners.top_left().y() > limit)
```
sz3, line 115: The check here is pretty simple. We run an extract on a freshly encoded image -- no blurring or skewing or any other camera-induced weirdness. We don't even need error correction for an image like this, it's pristine. The corners the scanner finds should be right where we expect them: at the edges of the image.

But if the (essentially random) data pattern is adversarial enough, the scanner will find a "corner" elsewhere, and we can detect/reject this frame as bad.

src/lib/encoder/test/EncoderTest.cpp
```diff
@@ -49,7 +49,7 @@ TEST_CASE( "EncoderTest/testFountain.4c", "[unit]" )
 	enc.set_legacy_mode();
 	assertEquals( 3, enc.encode_fountain(inputFile, outPrefix, 0) );
 
-	std::vector<uint64_t> hashes = {0xbb1cc62b662abfe5, 0xf586f6466a5b194, 0x8c2f0f40e6ecb08b};
+	std::vector<uint64_t> hashes = {0xbb1cc62b662abfe5, 0xf586f6466a5b194, 0x93a3830d042966e1};
```
sz3, line 52: Turns out we were hitting the error case here!

src/lib/encoder/Encoder.h
```diff
@@ -55,12 +56,27 @@ inline unsigned Encoder::encode_fountain(const std::string& filename, const std:
...
+		// will produce random patterns that falsely match as
+		// corner "anchors" and fail to extract. So:
+		// if frame fails the scan, skip it.
+		if (!Scanner::will_it_scan(*frame))
+		{
+			if (++consecutiveScansFailed < 5)
+				continue;
+
+			// else, we gotta make forward progress. And it's probably a bug?
+			std::cerr << fmt::format("generated {} bad frames in a row. This really shouldn't happen, maybe report a bug. :(", consecutiveScansFailed) << std::endl;
```
sz3, line 76: I expect that the likely way we hit this message is that the Scanner code gets messed up and everything (or many things) fail to scan. We could risk the infinite loop since that kind of failure should get caught by tests/etc, but... this seems smarter to me.

PR #121, opened 2025-03-01
make cimbar.exe work properly

wyjss:
fix compilation errors in the MSVC c++ environment and ensure the fountain mode codec works correctly.

### DIFF
--- a/src/third_party_lib/fmt/format.h
+++ b/src/third_party_lib/fmt/format.h
@@ -1453,11 +1453,7 @@ template <typename WChar, typename Buffer = memory_buffer> class to_utf8 {
         ++p;
         if (p == s.end() || (c & 0xfc00) != 0xd800 || (*p & 0xfc00) != 0xdc00) {
           if (policy == to_utf8_error_policy::abort) return false;
-#ifdef _WIN32
-          buf.append(string_view("\xef\xbf\xbd"));
-#else
           buf.append(string_view("�"));
-#endif
           --p;
         } else {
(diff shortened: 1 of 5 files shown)


src/third_party_lib/fmt/format.h
```diff
@@ -1453,7 +1453,11 @@ template <typename WChar, typename Buffer = memory_buffer> class to_utf8 {
         ++p;
         if (p == s.end() || (c & 0xfc00) != 0xd800 || (*p & 0xfc00) != 0xdc00) {
           if (policy == to_utf8_error_policy::abort) return false;
+#ifdef _WIN32
+          buf.append(string_view("\xef\xbf\xbd"));
```
sz3, line 1457: Sadly, this is not a change we can make here. Everything in `third_party_lib` is vendored from an upstream repo -- i.e. this file is fmt 10.1.0 verbatim.

It's worth noting that we're overdue for a dependency bump here (https://github.com/fmtlib/fmt is up to 11.1.4), but I'm not sure if that'll make this change or harder.

sz3:
Thanks for taking a swing at this -- not sure I can take it as is because of the fmtlib change, but it's useful regardless.

src/third_party_lib/fmt/format.h
wyjss: Got it, I have undone the non-conforming modifications. This issue exists only in fmt version 10.1.0.

sz3 (approved):

PR #124, opened 2025-04-03
opencv 4.11 for cimbar.js build + misc

sz3:
* small tweak to package-wasm script
* extra test case for fountain decoding (chasing a bug, but this doesn't seem to be it...)
* return an error code if no input files are specified in cimbar cli (seems good)
* some plumbing to  (eventually) allow different gl window sizes at runtime (i.e. swap barcode dimensions if different configs have different sizes)

This is preparation for merging a (separate) long-lived branch to allow more flexibility in config sizes.

### DIFF
--- a/package-wasm.sh
+++ b/package-wasm.sh
@@ -1,7 +1,8 @@
 #!/bin/bash
 #docker run --mount type=bind,source="$(pwd)",target="/usr/src/app" -it emscripten/emsdk:3.1.69 bash
 
-cd /usr/src/app
+CIMBAR_ROOT=${CIMBAR_ROOT:-/usr/src/app}
+cd $CIMBAR_ROOT
 
 apt update
 apt install python3 -y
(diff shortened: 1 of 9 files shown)


package-wasm.sh
```diff
@@ -1,29 +1,30 @@
 #!/bin/bash
 #docker run --mount type=bind,source="$(pwd)",target="/usr/src/app" -it emscripten/emsdk:3.1.69 bash
 
-cd /usr/src/app
+CIMBAR_ROOT=${CIMBAR_ROOT:-/usr/src/app}
```
sz3, line 4: Might switch this to dirname+readlink, but for now at least it's split out.

package-wasm.sh
```diff
@@ -1,29 +1,30 @@
...
+cd $CIMBAR_ROOT
 
 apt update
 apt install python3 -y
 
 cd opencv4/
 mkdir opencv-build-wasm
 cd opencv-build-wasm
-python3 ../platforms/js/build_js.py build_wasm --build_wasm --emscripten_dir=/emsdk/upstream/emscripten
+python3 ../platforms/js/build_js.py build_wasm --emscripten_dir=/emsdk/upstream/emscripten
```
sz3, line 13: `--build-wasm` is allegedly now the default.

src/lib/fountain/test/fountain_sinkSpecialTest.cpp
```diff
@@ -0,0 +1,93 @@
...
+}
+
+TEST_CASE( "FountainSinkSpecialTest/testMultipart", "[unit]" )
+{
+	::srand( ::time(nullptr) );
+	MakeTempDirectory tempdir;
+
+	fountain_decoder_sink<std::ofstream> sink(tempdir.path(), 750);
+
+	const int totalSize = 6000000;
```
sz3, line 63: The idea here is to permute through various random(ish) data permutations to see if anything trips up the decode. So far I've seen no issues.

src/lib/gui/window_glfw.h
```diff
@@ -127,7 +136,7 @@ class window_glfw
 	}
 
 protected:
-	GLFWwindow* _w;
+	GLFWwindow* _w = nullptr;
```
sz3, line 139: Just being prudent

PR #129, opened 2025-06-20
non-square grid dimensions + miscellany

sz3:
The longstanding "5x5" experiment (see the table here: https://github.com/sz3/libcimbar/issues/69#issuecomment-1512393960) led to various ideas, very few of which (sadly) really worked out.

But we have some code cleanup as a result, some bugfixes, and some open areas of experiment. Notably, the next version of the format (0.7.0? Perhaps?) will likely introduce a "widescreen" (something like 16:10) mode that targets 720p cameras in addition to the current 1:1 square modes.

+ update build scripts since ubuntu 20.04 is eol
+ better window scaling (ok, it didn't even have window scaling) for cimbar_send now that I have discovered the glfw docs

PR #130, opened 2025-06-20
Update default cimbar.js encoder config to "mode B" + wide config futureproofing

sz3:
A web UI specific followup to #129 

Also includes the minor UI improvement of hiding the hamburger menu if it overlaps the barcode.


web/main.js
```diff
@@ -42,7 +45,7 @@ function importFile(f)
 return {
   init : function(canvas)
   {
-    Module._initialize_GL(1040, 1040);
+    Main.setMode('B');
```
sz3, line 48: The new default. I introduced mode B and the autodetect decoder behavior a year ago+, but I didn't want to change the default before giving people time to upgrade.

(I may not push this to cimbar.org right away, but it'll be soon)

PR #133, opened 2025-07-31
Fix zstd skippable frame padding for small files

sz3:
This bug has been here from the start. :upside_down_face: 

*AFAIK* this doesn't manifest as anything, since the trailing padding seems to be reliably discarded (which is what we want). But the length calculation is *completely* wrong (for multiple kinds of wrong, in fact), so... let's fix it.

I *may* switch to `ZSTD_writeSkippableFrame()` for this, but with a proper test it's less compelling.

### DIFF
--- a/src/lib/encoder/test/EncoderRoundTripTest.cpp
+++ b/src/lib/encoder/test/EncoderRoundTripTest.cpp
@@ -31,7 +31,7 @@ TEST_CASE( "EncoderRoundTripTest/testFountain.Pad", "[unit]" )
 	Encoder enc(30, 4, 2);
 	assertEquals( 1, enc.encode_fountain(inputFile, outPrefix) );
 
-	uint64_t hash = 0xefcc8800efce8c48;
+	uint64_t hash = 0xeecc8800efce8808;
 	std::string path = fmt::format("{}_0.png", outPrefix);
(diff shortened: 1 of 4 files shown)


src/lib/encoder/test/EncoderRoundTripTest.cpp
```diff
@@ -67,7 +67,7 @@ TEST_CASE( "EncoderRoundTripTest/testFountain.SinkMismatch", "[unit]" )
 	enc.set_legacy_mode();
 	assertEquals( 1, enc.encode_fountain(inputFile, outPrefix) );
 
-	uint64_t hash = 0xaecc8c00efce8c28;
+	uint64_t hash = 0xaecc8808efce9c08;
```
sz3, line 70: since the padding changes, so does the output of these small file encodes :no_mouth:

web/main.js
```diff
@@ -24,11 +24,11 @@ function importFile(f)
 {
   const fileReader = new FileReader();
   fileReader.onload = (event) => {
-    const imageData = new Uint8Array(event.target.result);
-    const numBytes = imageData.length * imageData.BYTES_PER_ELEMENT;
+    const fileData = new Uint8Array(event.target.result);
```
sz3, line 27: unrelated js variable name cleanup

Issue #123, opened 2025-03-26
Decoding an animated cimbar on Linux.


owenkojola:
From looking at what documentation and code exists for libcimbar, cimbar, and cfc I was hoping to figure out an easy way to decode an animated cimbar displayed on a tv being captured by a webcam on a laptop. Does the decoder only take in png images or is it capable of taking a short video as input of an animated cimbar created with a fountain flag?

I really like the way the cfc app works but I didn't see anything about a PC version of that app I assume it doesn't exist yet based on reading other issue threads.

sz3:
I don't know about "easy", but `cimbar_recv` is the (test) program for this purpose. It's very (very) rough around the edges, and basically only exists for local dev-testing -- but if you can figure out the `/dev/` interface for the camera device it *could* be useful? (you might be able to extend it do what you want, maybe)

Another weirder option is to run the `cimbar` executable in its stdin mode (the default if you don't supply any input files) + have a separate video capture app dump images to /tmp (and probably clean them up every few seconds) + feed the /tmp filenames to `cimbar` line-by-line over stdin.

I'm hoping to improve the decoder situation in the next few months; I've made some time to work on it...

owenkojola:
I guess "easy" wasn't the correct term I should have used. "Easy" for a group of 4 people taking a senior design class for CS and IT bachelors. I was already thinking about doing the second option of just dumping pictures to a folder and letting it have at it. If you improve the decoder situation in the next few months, that would make things a lot easier, but I am sure I can figure it out without. Really nice work here I spent forever trying to find an alternative colored QR code/HCCB option that worked and wasn't just a "I have concepts of a plan" type of alternative.

sz3:
Webapp decoder is up at https://re.cimbar.org

(bits at https://github.com/sz3/libcimbar/releases/download/v0.6.2c/cimbar.wasm.tar.gz)

Issue #119, opened 2025-02-16
Feature Request: Python Bindings for libcimbar


TheRainstorm:
Hi

I'm a working on a project that involves generating and decoding color-icon-matrix barcodes. I came across libcimbar and was impressed by its performance and features. However, I noticed that there are no official Python bindings available, which makes it difficult for me to integrate it into my project. 

I believe that creating Python bindings for libcimbar would significantly broaden its usage. Thank you for your time and consideration.

Best regards,

sz3:
Been thinking about this. I think it could be a good thing to do, but only after I figure out the C API for decoding (there is one for encoding already). That'll mean it happens after the decoder work here (https://github.com/sz3/libcimbar/issues/71#issuecomment-2337147651)

Issue #104, opened 2024-09-12
Decoder web app


catcherxuefeng:
run decoder app in web

Anonymous3-a:
Scroll down on #71, it's got the discussion on that. The jist of it, though, is that quite a bit of work is needed.

Issue #100, opened 2024-09-05
Windows support


fcwys:
I hope GUI programs on the Windows platform can support encoding and decoding (from screenshots or cameras), which will greatly improve application scenarios and convenience.

zhjygit:
maybe, you can compile the code to exe.

fcwys:
The construction work is quite difficult for me, and I hope to have a Release version available for use. Thank you very much!

troilus:
You can use an electron converter to convert html to exe

fcwys:
The current HTML cannot send file names and does not support decoding, screenshot or camera recognition.

troilus:
About sending file name: https://github.com/sz3/libcimbar/issues/88
About decoding: https://github.com/sz3/libcimbar?tab=readme-ov-file#usage  maybe you can use some code to taking screenshots of screen, save and decode them.

sz3:
The dev tools do support decoding from images, and *could* be compiled for windows. But there are probably some patches needed and I'd need some help to get (and keep) things working, since it's not really a platform I use. That said, those are command line tools and may not be what you're looking for anyway.

The other path to windows support is via a (not yet existent) webapp decoder, which is something I intend to do, but will take some time.

fcwys:
It is a good idea to use a webapp decoder.

zhjygit:
> The dev tools do support decoding from images, and _could_ be compiled for windows. But there are probably some patches needed and I'd need some help to get (and keep) things working, since it's not really a platform I use. That said, those are command line tools and may not be what you're looking for anyway.
> 
> The other path to windows support is via a (not yet existent) webapp decoder, which is something I intend to do, but will take some time.

In windows platform,command tools is enough,we can encode and send to desktop window via scripts of command line.

sz3:
> Is there a decoder release version available on the Windows platform?

Not.

I've been making (slow...) progress on the webapp decoder, but it's still too buggy to be useful.

sz3:
https://github.com/sz3/libcimbar/releases/tag/v0.6.2c

PR #134, opened 2025-07-31
cimbar web decoder

sz3:
https://github.com/sz3/libcimbar/issues/104 , https://github.com/sz3/libcimbar/issues/100 , https://github.com/sz3/libcimbar/issues/99 , https://github.com/sz3/libcimbar/issues/71, https://github.com/sz3/libcimbar/issues/123 , ...

It's nearly time. The web decoder (mostly) works.

Expect this PR to sit for a bit -- I'll be taking some time to go over it again before the final merge, but it's such a massive diff that I think it'll be easier to review in chunks.

("review?" one may ask. "didn't you write the code?" one may follow up. Yes, I did, but that was my past self, and he is a somewhat different person.)

This too-large PR includes:
* a new set of c apis for decoding. These are prefixed with `cimbard_*`, and highly biased towards the JS/wasm implementation. However, they may also be useful for C bindings for other languages, eventually (https://github.com/sz3/libcimbar/issues/119)
     * specifically: multithreading needs drove the structure. The scan+extract+decode from image->raw bytes is one function (meant to be run statelessly on 2+ parallel threads), while the fountain decode functions are their own little world, run on a single thread to aggregate results. 
     * the Decoder bits have been split into filesystem-aware components and those without, as I did with the Encoder way back when. "SimpleEncoder" has been renamed to "Encoder", with "EncoderPlus" being the one with the extra FS methods.
     * the cimbar_recv executable has been updated (for now??) to use the new c apis
* renaming the c apis for encoding. These will now be prefixed with `cimbare_*`. They are mostly unchanged...
     * ... but the return codes have changed to be (perhaps) more sensible.
* a small 3rd set of c apis for zstd decompression. There are other ways to do this, including bringing in a JS dependency for zstd. However, we are already bundling/building zstd into our cimbar.js bits, so this didn't seem too crazy. It also gives a reasonable jumping off point for https://github.com/sz3/libcimbar/issues/88 , which will likely be a follow up PR.
* a new (separate for now) web page for the decoder, which will be at cimbar.org/recv.html (eventually but not yet). A beta version exists at https://re.cimbar.org , and hopefully that link will continue to work.
    * the "cimbar recv js" page will also be packaged with the wasm tar.gz, and can be served locally (e.g. with `python -m http.server` or `npx http-server` or ...) without an internet connection.
    * it can also be installed as a progressive web app (PWA) and run without an internet connection
    * ... unfortunately, until I'm more confident everything's working nicely I'm not going to combine the two apps into one. So it's two apps for a little while, one for sending and one for receiving.
    * crucially, the decoder app should work both for windows (https://github.com/sz3/libcimbar/issues/100) and for iphone (https://github.com/sz3/libcimbar/issues/99). Performance is better than expected, but slower than the native android app. I'll get actual numbers at some point.
    * ... there will surely be follow up PRs to fix bugs.
* a js test using the puppeteer+qunit. Forgive my choice of ancient js testing tech, *but* we need something that runs in a real browser.
    * currently only the recv component has automated tests. We'll extend that to send and the zstd stuff as well
    * there's a github workflow that runs the test.
 * misc tests. Will need to add more
 * added a wasm packaging script to rename all the files for cache busting
 * ... and other secrets! It's a big diff!

I wanted to split this into smaller pieces, but the coupling of the C apis to the decoder webapp is somewhat unavoidable, so we'll get it all in at once. There *will* be follow up PRs before the next release.


src/exe/cimbar_recv/recv.cpp
```diff
@@ -105,11 +110,12 @@ int main(int argc, char** argv)
...
-	Extractor ext;
-	Decoder dec(-1, -1);
+	// allocate buffers, etc
+	cimbard_configure_decode(colorBits, config_mode);
+	unsigned chunkSize = cimbar::Config::fountain_chunk_size(ecc, cimbar::Config::symbol_bits() + colorBits, legacy_mode);
 
-	unsigned chunkSize = cimbar::Config::fountain_chunk_size(ecc, colorBits+cimbar::Config::symbol_bits(), legacy_mode);
-	fountain_decoder_sink<cimbar::zstd_decompressor<std::ofstream>> sink(outpath, chunkSize);
+	std::vector<unsigned char> bufspace;
+	bufspace.resize(cimbard_get_bufsize(), 0);
```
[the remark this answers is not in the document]
sz3: Update: I'm thinking these changes will be somewhat temporary -- with a follow up PR, I'll fork the "cimbar_recv" into 2 separate testing apps, one that works the old way, and one that uses the new C api. The C api version (this one) will be the less full-featured one, but still useful for testing/troubleshooting.

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -41,12 +41,12 @@ int initialize_GL(int width, int height)
 	else
 		_window = std::make_shared<cimbar::window_glfw>(width, height, "Cimbar Encoder");
 	if (!_window or !_window->is_good())
-		return 0;
+		return -1;
 
-	return 1;
+	return 0;
```
sz3, line 46: Changing return values for more consistency with the new apis.

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -55,10 +55,24 @@ bool auto_scale_window()
 	return true;
 }
 
+// we may change the api to accept an buff to next_frame()
+// rather than generating a fresh cv::Mat alloc each time
+// but for now, for non-JS purposes we expose this function
```
sz3, line 60: .

src/lib/cimbar_js/cimbar_recv_js.cpp
```diff
@@ -0,0 +1,219 @@
...
+	TimeAccumulator _tScanExtract;
+	TimeAccumulator _tImgDecode;
+
+	// settings
+	unsigned _colorBits = 2;
+	int _modeVal = 68;
+
+	bool legacy_mode()
+	{
+		return _modeVal == 4;
```
sz3, line 34: Hoping to clean up this "legacy_mode" ugliness when I refactor the config system...

src/lib/cimbar_js/cimbar_recv_js.cpp
```diff
@@ -0,0 +1,219 @@
...
+			case 12:
+			{
+				img = cv::Mat(height * 3/2, width, CV_8UC1, imgdata).getUMat(cv::ACCESS_RW).clone();
+				cv::cvtColor(img, img, cv::COLOR_YUV2RGB_NV12); // 12 or 21 :hmm:
+				return img;
+			}
+			case 420:
+			{
+				img = cv::Mat(height * 3/2, width, CV_8UC1, imgdata).getUMat(cv::ACCESS_RW).clone();
+				cv::cvtColor(img, img, cv::COLOR_YUV420p2RGB);
```
sz3, line 51: Using firefox on linux, this does work for a claimed "I420" camera feed, but I'm not confident it's actually correct. Still investigating...

(worst case: we'll add a "just do the RGBA conversion up front" option in the UI)

src/lib/cimbar_js/cimbar_recv_js.h
```diff
@@ -0,0 +1,38 @@
...
+int64_t cimbard_fountain_decode(unsigned char* buffer, unsigned size);
+
+// get filesize from id
+int cimbard_get_filesize(uint32_t id);
+
+// if fountain_decode returned a >0 value, call this to retrieve the reassembled file
+// wherever a uint32_t id is passed, it should be on the same js thread
+// ... or at least in the same js shared memory...
+// as the fountain_decode() call
+int cimbard_finish_copy(uint32_t id, unsigned char* buffer, unsigned size);
```
sz3, line 30: The tests (and cimbar_recv) hopefully capture how these are used, but the gist is:

`cimbard_scan_extract_decode()` -> stateless, can be run in parallel (on different frames) on different threads. Stores decoded data in `bufspace`, and returns the size.

`cimbard_fountain_decode` -> stateful, should be run on the same thread. Aggregates results (from the scan_ex_dec call). The int64_t is a poor man's union, if it's negative it's a status code and if it's positive it's a uint32_t id we've succeeded, and the id can be used to access file state.

`cimbard_finish_copy` -> reassemble the completed transfer into buffer, using the id reported by `cimbard_fountain_decode`. *However*, these bytes are still (zstd) compressed, so there's an additional decompression step needed.

For decompression after `cimbard_finish_copy`, technically any zstd decompressor should work (probably). But the `cimbarz_*` functions are exposed for that purpose as well...

src/lib/cimbar_js/cimbar_zstd_js.h
```diff
@@ -0,0 +1,20 @@
...
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+int cimbarz_init_decompress(unsigned char* buffer, unsigned size);
+int cimbarz_get_filename(char* buffer, unsigned size);
+int cimbarz_get_bufsize();
+int cimbarz_decompress_read(unsigned char* buffer, unsigned size);
```
sz3, line 14: After `cimbard_finish_copy()` succeeds and we have our full zstd bytes reassembled, we call:
`cimbarz_init_decompress()` -> to set up the operation
`cimbarz_decompress_read()` -> until we don't have any more data to decompress

`cimbarz_get_filename` is a stub implementation, not implemented yet.

src/lib/cimbar_js/cimbar_zstd_js.cpp
```diff
@@ -0,0 +1,62 @@
...
+#include <sstream>
+
+// wasm api will be something like
+// zstd_init_stream(bytes) -> constructor, reads skippable frame at start? (needs to for ofstream case)
+// get_filename(buff, size) -> cached in wasmjs api from constructor call?
+// get_bytes(buff, size) -> "decompress_chunk()/write()"
+
+namespace {
+	// maybe a map eventually
+	std::unique_ptr<cimbar::zstd_decompressor<std::stringstream>> _dec;
```
sz3, line 16: I'm also thinking this should be a threadlocal? Perhaps? It's obnoxious to hold onto the state in general -- this is a compromise between various design choices around mapping a synchronous api (this one) to the async javascript "ReadableStream" interface. There are other options that might result in a cleaner C api, but cleaner is not always better imo (in general, I'd rather the C++ code take on the burden of jumping through more hoops than the JS does, since C++ is a more expressive language)

src/lib/cimbar_js/cimbar_zstd_js.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+}
+
+// zstd_decompressor constructor, reads skippable frame at start? (needs to for ofstream case)
+// if we make _dec a map, this'll also take the uin32_t id we got from the decoder sink, probably
+int cimbarz_init_decompress(unsigned char* buffer, unsigned size)
+{
+	if (_dec)
+		_dec.reset();
```
sz3, line 25: You can *try* to decompress multiple files in parallel, but you won't get very far...

src/lib/fountain/concurrent_fountain_decoder_sink.h
```diff
@@ -6,12 +6,11 @@
...
 #include <mutex>
 
-template <typename OUTSTREAM>
 class concurrent_fountain_decoder_sink
 {
 public:
-	concurrent_fountain_decoder_sink(std::string data_dir, unsigned chunk_size)
-	    : _decoder(data_dir, chunk_size)
+	concurrent_fountain_decoder_sink(unsigned chunk_size, const std::function<void(const std::string&, const std::vector<uint8_t>&)>& on_store=nullptr)
+		: _decoder(chunk_size, on_store)
```
sz3, line 13: Every time QtCreator messes with whitespace because my settings have subtly changed in the past 5 years, I get a little closer to adding clang-format to the project

web/index.html
```diff
@@ -1,302 +1,336 @@
 <!doctype html>
 <html lang="en" data-framework="javascript">
+
 <head>
-<title>Cimbar Encoder</title>
-<meta charset="utf-8">
-<meta name="viewport" content="width=device-width, user-scalable=no">
-<link rel="manifest" href="pwa.json">
+  <title>Cimbar Encoder</title>
```
sz3, line 5: Just formatting

web/main.js
```diff
@@ -1,275 +1,252 @@
...
-var _pause = 0;
+  var _interval = 66;
+  var _pause = 0;
 
-var _showStats = false;
-var _counter = 0;
-var _renderTime = 0;
+  var _showStats = false;
+  var _counter = 0;
+  var _renderTime = 0;
```
sz3, line 8: Also just formatting

web/pwa.json
```diff
@@ -1,7 +1,7 @@
 {
 	"lang": "en",
 	"dir": "ltr",
-	"name": "Cimbar Encoder",
+	"name": "cimbar.org Send",
```
sz3, line 4: Normalizing the naming a bit. The user facing tools will be called "send/receive", while the code itself will talk about things like encoding and decoding.

web/sw.js
```diff
@@ -1,5 +1,5 @@
 
-var _cacheName = 'cimbar-js-v0.6.1';
+var _cacheName = 'cimbar-js-vVERSION';
```
sz3, line 2: This'll be a date-based version string.

src/lib/compression/zstd_decompressor.h
```diff
@@ -82,12 +95,13 @@ class zstd_decompressor : public STREAM
...
 	{
 		return _lastError.str();
-	}
+	}	
 
 protected:
 	const size_t CHUNK_SIZE = ZSTD_DStreamInSize();
 	zstd_dstream _ds;
 	std::vector<char> _outBuff = std::vector<char>(ZSTD_DStreamOutSize());
+	std::string_view _inBuff;
```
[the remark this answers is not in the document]
sz3: The gist here is that that we want `write_once()` to stop once the output buffer is full. But we might not have worked through our input buffer all the way, so we need to "resume" on the next call.

It's probably not as bad as it feels to make this more stateful, since we already were holding onto a buffer (outBuff)

TraderWithPython:
This is a great project, thank you for your work. But I really can't help asking, when will the web decoder be officially launched?

sz3:
> This is a great project, thank you for your work. But I really can't help asking, when will the web decoder be officially launched?

Either sometime this week or next week, which will be the first (beta) release.

web/recv.js
```diff
@@ -0,0 +1,401 @@
+var Sink = function () {
```
sz3, line 1: I thought about doing file reassembly on a webworker -- and we may need to go that route at some point. We need the reassembled bytes (and maybe the zstd loop?) on the main thread for saving to disk, but there are a few different ways to get there. This is the simplest.

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+  var _fountainBuff = undefined;
+  var _errBuff = undefined;
+
+  // public interface
+  return {
+    on_decode: function (buff) {
+      if (_fountainBuff === undefined) {
+        const size = Module._cimbard_get_bufsize(); // max length of buff. We could also resize as we go...
+        const dataPtr = Module._malloc(size);
+        _fountainBuff = new Uint8Array(Module.HEAPU8.buffer, dataPtr, size);
```
sz3, line 12: reusing buffers. Technically this is a leak, since it will never get cleaned up -- but we should only do it once.

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+      }
+      if (buff.length == 0) { // sanity check
+        return;
+      }
+      _fountainBuff.set(buff);
+
+      console.log('sink decode ' + _fountainBuff); //TODO: base64?
```
sz3, line 19: Some of these log messages are pretty verbose. Will probably clean this up in a follow up PR

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+        else {
+          //Recv.download_bytes(buff, size + ".zst"); // size -> name, eventually
+          Zstd.decompress(id + "." + size, buff);
+        }
+      } catch (error) {
+        console.log("failed finish copy or download?? " + error);
+      }
+      // this needs to happen after decompress() completes
+      // currently decompress is sync, so it's fine. But...
+      Module._free(dataPtr);
```
sz3, line 76: ^

This is a bit of a diabolical (potential) race condition. It's easy to imagine this coming back to bite us in some future change, but it's hard to see what the right fix is until we hit the (future) problem.

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+      Recv.restart_video();
```
sz3, line 77: I'm not sure if we actually need this at this point. Leaving it in for now.

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+  var _recentDecode = -1;
+  var _recentExtract = -1;
+  var _renderTime = 0;
+  var _captureNextFrame = 0;
+
+  var _video = 0;
+  var _workers = [];
+  var _nextWorker = 0;
+  var _workerReady;
+  var _supportedFormats = ["NV12", "I420"]; // have cimbard_* return this somehow?
```
sz3, line 95: .

The translation step of these formats is somewhat awkward at the moment -- we create an arbitrary integer to map to these, and then duplicate the logic a few places. :shrug: 

Perhaps a more interesting thing to note is that these are formats that we do the format conversion for on the worker thread, not on the main thread. I'd like to get some samples of I422, I444, to add them to here (see full list: https://developer.mozilla.org/en-US/docs/Web/API/VideoFrame/format), perhaps using the "capture" option from the menu.

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+    restart_video: function () {
+      console.log("restart video?");
+      if (_video) {
+        return;
+      }
+      if (!('srcObject' in _video)) {
+        console.log("can't restart video? :|")
+        return;
+      }
+      // ios cancels the feed sometimes. Restart it...?
```
sz3, line 194: I mentioned this above, but I don't think we necessarily need this anymore. I think this case may have been due to an unnecessary `alert()`

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+      setTimeout(function () {
+        return window.URL.revokeObjectURL(bloburl);
+      }, 1000);
+    },
+
+    on_decode: function (wid, data) {
+      console.log('Main thread received message from worker' + wid + ':', data);
+      // if extract but no bytes, log extract counter
+      if (data.nodata) {
+        _recentExtract = _counter;
```
sz3, line 214: not returning here is a (mostly harmless) bug -- will fix in a followup PR

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+      // https://developer.mozilla.org/en-US/docs/Web/API/VideoFrame
+
+      _counter += 1;
+      if (_workers.length == 0)
+        return;
+      if (_nextWorker >= _workers.length)
+        _nextWorker = 0;
+
+      // piggyback off this call to make sure our visual state is correct
+      Recv.update_visual_state();
```
sz3, line 245: we're calling this here because this is the function that runs on an interval. If we wanted to, we could do setInterval(update_visual_state) and not worry about it here. :thinking:

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+        console.log(current.length);
+      }
+      for (var i = 0; i < report.length; i++) {
+        console.log(report[i] * 100 + "%");
+        current[i].style.width = report[i] * 100 + "%";
+      }
+    },
+
+    toggleFullscreen: function () {
+      // nothin yet
```
sz3, line 348: follow up PR

web/zstd.js
```diff
@@ -0,0 +1,116 @@
...
+    const response = new Response(readstream);
+    const blob = await response.blob();
+    Zstd.download_blob(filename, blob);
+  }
+
+  function saveToFile(readstream, filename) {
+    ///if (window.showSaveFilePicker == undefined)
+    saveViaBlob(readstream, filename);
+    //else
+    //  saveViaStreaming(readstream, filename);
```
sz3, line 87: infuriatingly, chrome-based browsers (the only ones that support this showSaveFilePicker functionality) require a user action for the save dialog to activate. But we don't technically have one of those, since "hold the camera in place for 10 seconds" doesn't count.

We can sort of force it with `alert()`, but if we do that we get two dialog popups (alert -> save) instead of one (save).

Anyway, the showSaveFilePicker way is here if we need it.

src/lib/cimbar_js/cimbar_zstd_js.cpp
```diff
@@ -0,0 +1,62 @@
...
+
+// "decompress_chunk()/write()"
+int cimbarz_decompress_read(unsigned char* buffer, unsigned size)
+{
+	if (!_dec)
+		return -1;
+	if (!_dec->good())
+		return -2;
+
+	_dec->str(std::string());
```
sz3, line 55: This is how you reset the buffer, due to strange stringstream reasons.

.github/workflows/wasm.yml
```diff
@@ -0,0 +1,62 @@
+name: wasm
+
+on: push
```
sz3, line 3: Might make this PR-only. It feels excessive to run every time.

.gitignore
```diff
@@ -4,6 +4,7 @@ Testing/
 build*/
 dist/
 opencv4/
+web/test/node_modules/
```
sz3, line 7: only for testing

package-wasm.sh
```diff
@@ -17,7 +18,11 @@ mkdir build-wasm
 cd build-wasm
 emcmake cmake .. -DUSE_WASM=1 -DOPENCV_DIR=$CIMBAR_ROOT/opencv4
 make -j5 install
-(cd ../web/ && tar -czvf cimbar.wasm.tar.gz cimbar_js.js cimbar_js.wasm index.html main.js)
+(cd ../web/ && bash wasmgz.sh)
+
+if [ -n $SKIP_JS ]; then
```
sz3, line 23: This is mostly for the automated testing case. But it can be useful for dev as well.

src/exe/cimbar/cimbar.cpp
```diff
@@ -282,12 +282,12 @@ int main(int argc, char** argv)
...
 	if (compressionLevel <= 0)
 	{
-		fountain_decoder_sink<std::ofstream> sink(outpath, chunkSize, true);
+		fountain_decoder_sink sink(chunkSize, write_on_store<std::ofstream>(outpath, true));
 		res = decode(infiles, fountain_decode_fun(sink, d), no_deskew, undistort, color_mode, preprocess, color_correct);
 	}
 	else // default case, all bells and whistles
 	{
-		fountain_decoder_sink<cimbar::zstd_decompressor<std::ofstream>> sink(outpath, chunkSize, true);
+		fountain_decoder_sink sink(chunkSize, write_on_store<cimbar::zstd_decompressor<std::ofstream>>(outpath, true));
```
sz3, line 290: With decompression.

src/exe/cimbar_recv/recv.cpp
```diff
@@ -130,29 +136,41 @@ int main(int argc, char** argv)
...
-		if (!res)
+		// extract, decode, etc
+		int bytes = cimbard_scan_extract_decode(img.data, img.cols, img.rows, 3, bufspace.data(), bufspace.size());
+		if (bytes <= 0)
+			continue;
+
+		if (bytes % chunkSize != 0)
 		{
-			//std::cerr << "no extract " << mat.cols << "," << mat.rows << std::endl;
+			std::cerr << "WEIRD SEEMS BAD? " << count << std::endl;
```
sz3, line 152: I don't think this is actually possible. We'll remove it when we have more confidence.

src/exe/cimbar_recv/recv.cpp
```diff
@@ -130,29 +136,41 @@ int main(int argc, char** argv)
...
-			std::cerr << "got some bytes " << bytes << std::endl;
+		int64_t res = cimbard_fountain_decode(bufspace.data(), bytes);
+		if (res > 0)
+		{
+			// attempt save
+			uint32_t fileId = res;
 
-		std::cerr << turbo::str::join(sink.get_progress()) << std::endl;
+			unsigned size = cimbard_get_filesize(fileId);
+			std::string filename = fmt::format("0.{}", size);
```
sz3, line 163: We should *probably* should pull the encode_id out of the fileId for this filename. :thinking: 

(actual filename support is coming)

src/exe/cimbar_send/send.cpp
```diff
@@ -77,13 +77,13 @@ int main(int argc, char** argv)
...
-	if (!initialize_GL(window_size_x, window_size_y))
+	if (cimbare_init_window(window_size_x, window_size_y) < 0)
 	{
 		std::cerr << "failed to create window :(" << std::endl;
 		return 70;
 	}
-	auto_scale_window();
-	configure(colorBits, ecc, compressionLevel, legacy_mode);
+	cimbare_auto_scale_window();
+	cimbare_configure(colorBits, ecc, compressionLevel, legacy_mode);
```
sz3, line 86: I'm not sure what to do about the split `*_configure()` calls. It's tempting to smash them together, and that's probably the end state. But I need to think about it more.

src/lib/cimb_translator/GridConf.h
```diff
@@ -64,17 +64,17 @@ namespace cimbar
...
 		static constexpr unsigned color_bits = 2;
 		static constexpr unsigned symbol_bits = 4;
-		static constexpr unsigned ecc_bytes = 35;
-		static constexpr unsigned ecc_block_size = 174;
-		static constexpr unsigned image_size_x = 1285;
-		static constexpr unsigned image_size_y = 736;
+		static constexpr unsigned ecc_bytes = 36;
+		static constexpr unsigned ecc_block_size = 179;
+		static constexpr unsigned image_size_x = 1024;
+		static constexpr unsigned image_size_y = 720;
```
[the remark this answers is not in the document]
sz3: It also helps for cheap 1080p cameras that suck at autofocus.

src/lib/cimbar_js/CMakeLists.txt
```diff
@@ -46,9 +51,10 @@ set (LINK_WASM_LIST
 	-s MIN_WEBGL_VERSION=2
 	-s MAX_WEBGL_VERSION=2
 	-s USE_GLFW=3
+	-s WASM_BIGINT=1
 	-s FILESYSTEM=0
 	-s TOTAL_MEMORY=134217728
-	-s EXPORTED_FUNCTIONS='["_render","_next_frame","_encode","_configure","_get_aspect_ratio","_malloc","_free"]'
+	-s EXPORTED_FUNCTIONS='["_cimbare_render","_cimbare_next_frame","_cimbare_encode","_cimbare_configure","_cimbare_get_aspect_ratio","_cimbard_get_report","_cimbard_scan_extract_decode","_cimbard_fountain_decode","_cimbard_get_bufsize","_cimbard_get_filesize","_cimbard_finish_copy","_cimbard_configure_decode","_cimbarz_init_decompress","_cimbarz_get_bufsize","_cimbarz_decompress_read","_malloc","_free"]'
```
sz3, line 57: I tried to split this line but clang got mad :scream: 

Maybe some `\`s would help?

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -55,10 +55,24 @@ bool auto_scale_window()
...
+	if (_next->cols == 0 or _next->rows == 0)
+		return -1;
+
+	*buff = _next->data;
+	return _next->cols * _next->rows * _next->channels();
+}
+
 // render() and next_frame() could be put in the same function,
 // but it seems cleaner to split them.
 // in any case, we're concerned with frame pacing (some encodes take longer than others)
```
sz3, line 74: I *think* this comment is less relevant as the scope of this api slowly grows. Having them separate is clearly the right way imo

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -85,26 +99,30 @@ int next_frame()
 	if (_fes->block_count() > required)
 	{
 		_fes->restart();
-		_window->shake(0);
+		if (_window)
```
sz3, line 102: We might want to use these apis without creating a window.

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -136,7 +154,9 @@ int configure(unsigned color_bits, unsigned ecc, int compression, bool legacy_mo
 	// make sure we've initialized
 	int window_size_x = cimbar::Config::image_size_x() + 16;
 	int window_size_y = cimbar::Config::image_size_y() + 16;
-	initialize_GL(window_size_x, window_size_y);
+	int initRes = cimbare_init_window(window_size_x, window_size_y);
+	if (initRes < 0)
+		return initRes;
```
sz3, line 159: Good to know if it blows up in our face. This function is how the JS interacts with init_window.

src/lib/cimbar_js/cimbar_recv_js.cpp
```diff
@@ -0,0 +1,219 @@
...
+	std::string _reporting;
+
+	cv::Mat _debugFrame;
+
+	TimeAccumulator _tScanExtract;
+	TimeAccumulator _tImgDecode;
+
+	// settings
+	unsigned _colorBits = 2;
+	int _modeVal = 68;
```
sz3, line 30: The magic number spreads...

src/lib/cimbar_js/cimbar_recv_js.cpp
sz3: Was able to get a I420 out of an android phone that matches. Still wondering if the "force RGBA" option might be a good idea.

src/lib/cimbar_js/cimbar_recv_js.cpp
```diff
@@ -0,0 +1,219 @@
...
+			{
+				img = cv::Mat(height * 3/2, width, CV_8UC1, imgdata).getUMat(cv::ACCESS_RW).clone();
+				cv::cvtColor(img, img, cv::COLOR_YUV420p2RGB);
+				return img;
+			}
+			default:
+				break;
+		}
+
+		int cvtype = type==4? CV_8UC4 : CV_8UC3;
```
sz3, line 58: implicit case 3+4 here

src/lib/cimbar_js/cimbar_recv_js.cpp
```diff
@@ -0,0 +1,219 @@
...
+				cimbar::Config::ecc_bytes(),
+				cimbar::Config::symbol_bits() + _colorBits,
+				legacy_mode()
+	);
+	// early bail if bufsize doesn't match config params (fountain chunk size * count)
+	if (bufsize < chunkSize * chunksPerFrame)
+		return -2;
+
+	// interface to take the aligned output buffers of chunkSize and dump them into bufspace
+	escrow_buffer_writer ebw(bufspace, chunksPerFrame, chunkSize);
```
sz3, line 115: This feels slightly awkward, but the gist is that we know there's a hard upper bound on the size of an output from a call to `decode_fountain()` -- and we also know we're only making one call here. So we can have this object (which functions a bit like an fstream) hold the output buffer memory location, which allows the decoded bytes to land exactly where we need them.

src/lib/cimbar_js/cimbar_recv_js.cpp
```diff
@@ -0,0 +1,219 @@
...
+	if (mode_val <= 0)
+		mode_val = 68;
+
+	bool refresh = (color_bits != _colorBits or mode_val != _modeVal);
+	if (refresh)
+	{
+		// update config
+		_colorBits = color_bits;
+		_modeVal = mode_val;
+		_sink.reset();
```
sz3, line 213: Leaning on the lazy create here.

src/lib/compression/zstd_decompressor.h
```diff
@@ -18,18 +18,15 @@ class zstd_decompressor : public STREAM
 	using STREAM::STREAM; // pull in constructors
 
 public:
-	bool write(const char* data, size_t len)
+	bool write_once()
```
sz3, line 21: Basically this is just a loop unrolling. As a result, we need to hold on to `_inBuff` for longer.

This is potentially a bad idea, since we don't own that memory. I don't love it...

src/lib/encoder/test/escrow_buffer_writerTest.cpp
```diff
@@ -0,0 +1,61 @@
...
+{
+	// the way we expect to use this class is with the aligner wrapping it
+	std::vector<unsigned char> bufspace;
+	bufspace.resize(50, 0);
+
+	escrow_buffer_writer ebw(bufspace.data(), 5, 10); // two params should == overall buffer size
+	aligned_stream aligner(ebw, 10, 0);
+	assertTrue( aligner.good() );
+
+	aligner.write("01234567890123456789012345", 26);
```
sz3, line 47: We write in whatever weird increment we want, and the aligner flushes it when the full chunk is ready.

src/lib/extractor/test/ExtractorTest.cpp
```diff
@@ -14,7 +14,7 @@ TEST_CASE( "ExtractorTest/testExtract", "[unit]" )
 	MakeTempDirectory tempdir;
 
 	std::string imgPath = tempdir.path() / "ex.jpg";
-	Extractor ext({1024, 1024}, 30);
+	ExtractorPlus ext({1024, 1024}, 30);
```
sz3, line 17: Extractor -> ExtractorPlus

src/lib/fountain/FountainDecoder.h
```diff
@@ -46,21 +46,31 @@ class FountainDecoder
 		return _res;
 	}
 
-	std::optional<std::vector<uint8_t>> decode(unsigned block_num, uint8_t* data, size_t length)
+	bool decode(unsigned block_num, uint8_t* data, size_t length)
 	{
 		auto pear = _seenBlocks.insert(block_num);
 		if (!pear.second)
-			return std::nullopt;
+			return false;
```
sz3, line 53: My ideal api here is that we'd call `write()`, and trigger write->recover->decompress->disk write.

For the first pass, I didn't want to hide all the individual steps. I might put it back later and clean up the api.

With the added knowledge of the zstd_decompressor inBuff situation (see whining above), I think having the caller own the wirehair_recover buffer might be a mistake. And if we need to own the memory here anyway, that swings the ideal implementation (as I judge it) back to how it was before these changes. So.. stay tuned

src/lib/fountain/concurrent_fountain_decoder_sink.h
sz3: (I've been planning on doing it. Though I hate what they do to my precious anonymous namespaces)

src/lib/fountain/fountain_decoder_sink.h
```diff
@@ -33,13 +45,21 @@ class fountain_decoder_sink
...
+
+	bool store(const FountainMetadata& md, fountain_decoder_stream& s)
+	{
+		if (_onStore)
+		{
+			auto res = s.recover();
+			if (!res)
+				return false;
+			_onStore(get_filename(md), *res);
+			mark_done(md);
```
sz3, line 61: Notably, caller *must* explicitly call `mark_done()` if the function callback isn't in use here.

src/lib/fountain/fountain_decoder_sink.h
```diff
@@ -86,60 +106,79 @@ class fountain_decoder_sink
...
-
-		if (store(md, *finished))
-			mark_done(md);
-		return true;
+			return 0;
+
+		// when you provide a write callback,
+		// store() will call mark_done() afterwards
+		// -- and the assembled file will we dropped from RAM.
+		// but if no callback is provided, you can do something else.
```
sz3, line 139: ^ and this is why mark_done() is important...

src/lib/util/Timer.h
```diff
@@ -3,20 +3,49 @@
 
 #include <ctime>
 
+class TimeAccumulator
```
sz3, line 6: Making timing slightly easier. Will switch CFC to use this I think.

web/recv-worker.js
```diff
@@ -0,0 +1,134 @@
...
+      var type = 4;
+      if (format == "NV12") {
+        type = 12;
+      }
+      else if (format == "I420") {
+        type = 420;
+      }
+
+      try {
+        // then decode in wasm, fool
```
sz3, line 41: I'm the fool

web/recv-worker.js
```diff
@@ -0,0 +1,134 @@
...
+          }
+        }
+        else { //if (len > 0) {
+          console.log('len is ' + len);
+          //self.postMessage({ error: true, res: len });
+          const msgbuf = new Uint8Array(Module.HEAPU8.buffer, fountainBuff.byteOffset, len).slice();
+          console.log(msgbuf);
+          self.postMessage(msgbuf.buffer, [msgbuf.buffer]);
+        }
+        // in main, const receivedArray = new Uint8Array(event.data);
```
sz3, line 63: .

web/recv-worker.js
```diff
@@ -0,0 +1,134 @@
...
+    },
+
+    mallocPlease: function (name, size) {
+      if (_buffs[name] === undefined || size > _buffs[name].length) {
+        try {
+          if (size > _buffs[name].length) {
+            console.log("resizing " + name + " buff from " + _buffs[name].length + " to " + size);
+            Module._free(_buffs[name].byteOffset);
+          }
+        } catch (e) { } // if we're leaking memory we'll find out the hard way
```
sz3, line 88: *probably* should log a message here :upside_down_face:

web/recv-worker.js
```diff
@@ -0,0 +1,134 @@
...
+  if (!_wasmInitialized) {
+    console.log('we got no wasm :(');
+    self.postMessage({ type: 'startWasm', error: "no wasm" });
+    return;
+  }
+
+  if (event.data.config) {
+    console.log('attempting config');
+    try {
+      Module._cimbard_configure_decode(event.data.color_bits, event.data.mode_val);
```
sz3, line 125: Each wasm thread -- right now (since we're not using shared memory buffers) -- is essentially its own process space. Meaning globals are actually (basically) threadlocals. Meaning the config has to be run independently on each worker.

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+      try {
+        vf = new VideoFrame(_video, { timestamp: now });
+        const width = vf.displayWidth;
+        const height = vf.displayHeight;
+        Recv.set_HTML("errorbox", vf.format, true);
+
+        // try to use the default format, but only if we can decode it...
+        let vfparams = {};
+        if (!_supportedFormats.includes(vf.format)) {
+          vfparams.format = "RGBA";
```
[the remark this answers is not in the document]
sz3: But I've also been thinking that a user option to force conversion to RGBA here might be handy.

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+        }
+        const size = vf.allocationSize(vfparams);
+        const buff = new Uint8Array(size);
+        vf.copyTo(buff, vfparams);
+
+        let format = vfparams.format || vf.format;
+        if (format == "RGBA" && size != width * height * 4) {
+          format = vf.format; //fallback
```
sz3, line 265: It turns out that on ios, you can ask it for RGBA and it'll happily give you back NV12.

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+    },
+
+    captureFrame: function () {
+      _captureNextFrame = 1;
+      alert("about to capture!");
+    },
+
+    download_bytes: function (buff, name) {
+      var blob = new Blob([buff], { type: 'application/octet-stream' });
+      Zstd.download_blob(name, blob);
```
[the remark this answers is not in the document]
sz3: edit: I decided, look forward to the PR

web/recv.js
```diff
@@ -0,0 +1,401 @@
...
+    },
+
+    update_visual_state: function () {
+      // check counters
+      var xh1 = document.getElementById("crosshair1");
+      var xh2 = document.getElementById("crosshair2");
+      if (_recentDecode > 0 && _recentDecode + 30 > _counter) {
```
sz3, line 297: This is to give us some visual feedback on in-progress transfers.

web/sw-recv.js
```diff
@@ -0,0 +1,43 @@
+
+var _cacheName = 'cimbar-recv-js-vVERSION';
+var _cacheFiles = [
```
sz3, line 3: I'm still concerned whether I'm doing this service worker caching stuff right. We'll see :grimacing:

web/test/run-in-browser.js
```diff
@@ -0,0 +1,55 @@
+// simple test scaffolding to run qunit in browser
+// using puppeteer for now, playwright might be better???
+// copy+paste, ai cross reference, I didn't write this,
+// run at your own risk etc (it's pretty simple though)
```
sz3, line 4: Gross AI-synthesized tutorial code :grimacing: 

At least it's simple enough.

web/test/test_base.js
```diff
@@ -0,0 +1,109 @@
...
+    calls: [],
+
+    decompress: function (name, data) {
+      Zstd.calls.push({ decompress: [name, data.length] });
+    }
+  };
+}();
+
+
+function wait_for(assert, block) {
```
sz3, line 18: Every UI testing library ends up needing this function

web/test/test_base.js
```diff
@@ -0,0 +1,109 @@
...
+  });
+}
+
+async function load_image(num) {
+  const img = document.getElementById('example_frame' + num);
+  const imageBitmap = await window.createImageBitmap(img);
+  imageBitmap.requestVideoFrameCallback = function () { };
+
+  try {
+    Recv.init_video(imageBitmap);
```
sz3, line 49: Very helpfully, the call we're using to extract information from a videoframe (which is where the camera feed ends up) also works perfectly well for an imagebitmap. So for our test, we can give it a few in succession to imitate pulling images from the camera.

web/wasmgz.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+# run from within web dir
+# does various renames for cache busting
+
+RENAME_FILES=$(ls cimbar_js.js cimbar_js.wasm main.js recv*.js zstd.js sw*.js pwa*.json)
```
sz3, line 5: Could've written this in python. But bash is old reliable

web/zstd.js
```diff
@@ -0,0 +1,116 @@
...
+      },
+      pull(controller) {
+        const bytesRead = Module._cimbarz_decompress_read(_decompBuff, bufferSize);
+        console.log("reader got " + bytesRead);
+        if (bytesRead <= 0) {
+          // No more data, close the stream
+          controller.close();
+        } else {
+          // Copy the data from WASM memory to a JavaScript Uint8Array
+          const chunk = new Uint8Array(Module.HEAPU8.buffer, _decompBuff, bytesRead).slice();
```
sz3, line 42: slice() == copy

Issue #88, opened 2024-02-07
send file name


notune:
I think it would be a great feature to include the file name as well. (or maybe just the file extension). This would only require a few additional bytes to be sent.

sz3:
It'd be helpful.

I left it out initially due to (perceived) security risk on my part. I've sort of come around on it though -- I don't think it's *too* scary to have the file declare what it is, since a troublemaker would be able to do the same sort of things without it anyway:

> :smiling_imp: : just rename 1.123456 to "notavirus.pdf" and open it!

vs just getting the save dialogue for "notavirus.pdf"

There's kind of an inherent trust issue with being able to send *the whole file* in the bar code regardless, so I'd like to try to communicate that better as I (slowly) add more usability.

sz3:
Implementation wise, I alluded to it [here](https://github.com/sz3/libcimbar/issues/77#issuecomment-1722308518)

> another option is to use the "skippable frame" functionality of zstd to attach some metadata to the transfer.

That's probably how I'd do it: make it entirely optional on the client's end as to whether to pay attention to the filename, but make sure the sender includes it.

We're already using skippable frames to pad small files up to the fountain (wirehair) block size, so it's not *too* crazy.

sz3:
As of the 0.6.2 release (https://github.com/sz3/libcimbar/releases/tag/v0.6.2c), this should work! ... hopefully!

The android app release should follow soon.

PR #135, opened 2025-08-05
Include filename when sending/encoding a file + use it when receiving/decoding

sz3:
For https://github.com/sz3/libcimbar/issues/88 .

The idea is that we have a skippable zstd frame with metadata at the start of the encoded file, and check for it when decoding. If the metadata frame fails to decode, or doesn't exist (as will be the case for any encoder <0.6.2), we fall back to the previous behavior.

Follow up to #134 .

### DIFF
--- a/src/exe/cimbar_recv/recv.cpp
+++ b/src/exe/cimbar_recv/recv.cpp
@@ -169,7 +169,7 @@ int main(int argc, char** argv)
 			if (res != 0)
 				std::cerr << "failed fountain_finish_copy " << res << std::endl;
 
-			write_on_store<cimbar::zstd_decompressor<std::ofstream>>(outpath, true)(filename, data);
+			decompress_on_store<std::ofstream>(outpath, true)(filename, data);
 		}
 	}
 
--- a/src/lib/cimbar_js/cimbar_js.cpp
+++ b/src/lib/cimbar_js/cimbar_js.cpp
@@ -113,7 +113,7 @@ int cimbare_next_frame()
 	return ++_frameCount;
 }
 
-int cimbare_encode(unsigned char* buffer, unsigned size, int encode_id)
+int cimbare_encode(const unsigned char* buffer, unsigned size, const char* filename, unsigned fnsize, int encode_id)
 {
 	_frameCount = 0;
 	if (!FountainInit::init())
@@ -131,8 +131,8 @@ int cimbare_encode(unsigned char* buffer, unsigned size, int encode_id)
 	else
 		enc.set_encode_id(static_cast<uint8_t>(encode_id));
 
-	cimbar::byte_istream bis(reinterpret_cast<char*>(buffer), size);
-	_fes = enc.create_fountain_encoder(bis, _compressionLevel);
+	cimbar::byte_istream bis(reinterpret_cast<const char*>(buffer), size);
+	_fes = enc.create_fountain_encoder(bis, std::string_view(filename, fnsize), _compressionLevel);
 
 	if (!_fes)
 		return -1; // return -1 plz
--- a/src/lib/cimbar_js/cimbar_recv_js.cpp
+++ b/src/lib/cimbar_js/cimbar_recv_js.cpp
@@ -2,6 +2,7 @@
 #include "cimbar_recv_js.h"
 
 #include "cimb_translator/Config.h"
+#include "compression/zstd_header_check.h"
 #include "encoder/Decoder.h"
 #include "encoder/escrow_buffer_writer.h"
 #include "extractor/Extractor.h"
@@ -105,7 +106,7 @@ int cimbard_get_bufsize()
 	return fountain_chunks_per_frame() * fountain_chunk_size();
 }
 
-int cimbard_scan_extract_decode(uchar* imgdata, unsigned imgw, unsigned imgh, int format, uchar* bufspace, unsigned bufsize)
+int cimbard_scan_extract_decode(const uchar* imgdata, unsigned imgw, unsigned imgh, int format, uchar* bufspace, unsigned bufsize)
 {
 	if (format <= 0)
 		format = 3;
@@ -149,7 +150,7 @@ int cimbard_scan_extract_decode(uchar* imgdata, unsigned imgw, unsigned imgh, in
 }
 
 // returns id of final file (can be used to get size of `finish_copy`'s buffer) if complete, 0 if success, -1 on error
-int64_t cimbard_fountain_decode(unsigned char* buffer, unsigned size)
+int64_t cimbard_fountain_decode(const unsigned char* buffer, unsigned size)
 {
 	unsigned chunkSize = fountain_chunk_size();
 	if (!_sink) // lazy-create the sink on first run
@@ -163,7 +164,7 @@ int64_t cimbard_fountain_decode(unsigned char* buffer, unsigned size)
 	{
 		/*std::cout << fmt::format("buff {} of {} -- {},{},{},{},{},{}", i, size, (unsigned)buffer[0+i], (unsigned)buffer[1+i],
 				(unsigned)buffer[2+i], (unsigned)buffer[3+i], (unsigned)buffer[4+i], (unsigned)buffer[5+i]) << std::endl;*/
-		res = _sink->decode_frame(reinterpret_cast<char*>(buffer+i), chunkSize);
+		res = _sink->decode_frame(reinterpret_cast<const char*>(buffer+i), chunkSize);
 	}
 
 	std::cout << "fountain decode res is " << res << std::endl;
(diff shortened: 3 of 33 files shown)


src/exe/cimbar_recv/recv.cpp
```diff
@@ -136,41 +130,29 @@ int main(int argc, char** argv)
...
 
-		// extract, decode, etc
-		int bytes = cimbard_scan_extract_decode(img.data, img.cols, img.rows, 3, bufspace.data(), bufspace.size());
-		if (bytes <= 0)
-			continue;
-
-		if (bytes % chunkSize != 0)
+		// extract
+		bool shouldPreprocess = false;
+		int res = ext.extract(img, img);
```
sz3, line 141: In #134, I updated the cimbar_recv app to use the new C api rather than the C++ classes directly. The android app uses the C++ classes though, and probably will continue to do so. So... for now, we'll just have 2 different test apps. cimbar_recv2 will use the C api, cimbar_recv the C++ classes.

sz3: I do have some hope that at some point I'll be happy enough with the C api to switch to using it everywhere, but it's definitely not there yet.

src/lib/cimbar_js/cimbar_recv_js.h
```diff
@@ -14,11 +14,11 @@ unsigned cimbard_get_debug(unsigned char* buff, unsigned maxlen);
 // imgsize=width*height*channels for rgba. Other formats are weirder.
 // output of scan is stored in `bufspace`
 int cimbard_get_bufsize();
-int cimbard_scan_extract_decode(unsigned char* imgdata, unsigned imgw, unsigned imgh, int format, unsigned char* bufspace, unsigned bufsize);
+int cimbard_scan_extract_decode(const unsigned char* imgdata, unsigned imgw, unsigned imgh, int format, unsigned char* bufspace, unsigned bufsize);
```
sz3, line 17: const!

src/lib/compression/zstd_compressor.h
```diff
@@ -72,23 +72,27 @@ class zstd_compressor : public STREAM
...
 		if (len < 9)
 			len = 9;
 
-		len -= 8;
-		std::array<char, 8> header = {0x50, 0x2A, 0x4D, 0x18, (char)(len&0xFF), (char)((len&0xFF00) >> 8), (char)((len&0xFF0000) >> 16), 0};
-		STREAM::write(header.data(), header.size());
+		// test what this actually does with ZSTD_isSkippableFrame and ZSTD_readSkippableFrame
+		// then, replace it with ZSTD_writeSkippableFrame...
+		std::string temp(len-8, '\0');
+		size_t writ = ZSTD_writeSkippableFrame(_compBuff.data(), _compBuff.size(), temp.data(), temp.size(), 0);
```
sz3, line 83: This is the "unstable" static zstd api, which I think is why I was hesitant to use it initially. But it doesn't seem to have changed in 5+ years, so it's probably fine

src/lib/fountain/fountain_decoder_sink.h
```diff
@@ -23,13 +24,35 @@ std::function<void(const std::string&, const std::vector<uint8_t>&)> write_on_st
...
+		return filename;
 	};
 }
 
+template <typename OUTSTREAM>
+std::function<std::string(const std::string&, const std::vector<uint8_t>&)> decompress_on_store(std::string data_dir, bool log_writes=false)
+{
+	return [data_dir, log_writes](const std::string& fallback_name, const std::vector<uint8_t>& data)
+	{
+		std::string filename = cimbar::zstd_header_check::get_filename(data.data(), data.size());
```
sz3, line 36: We could attack this a couple ways:
1. when decompressing, read everything once, and check the opening read for our metadata. This would require lazy opening the file when writing. (e.g. we'd have to open on write(), not open())
2. read the opening bytes twice. Once here, and again in the decompressor.

The key insight here (imo) is that the input is already in RAM. So reading twice doesn't hurt that much, and maintaining a semblance of consistency in the decompressor api makes it a reasonable tradeoff. 

So we're doing (2)

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -113,7 +113,7 @@ int cimbare_next_frame()
 	return ++_frameCount;
 }
 
-int cimbare_encode(unsigned char* buffer, unsigned size, int encode_id)
+int cimbare_encode(const unsigned char* buffer, unsigned size, const char* filename, unsigned fnsize, int encode_id)
```
sz3, line 116: The api should communicate that these are const (since they *are*, whether they're using the keyword or not).

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -131,8 +131,8 @@ int cimbare_encode(unsigned char* buffer, unsigned size, int encode_id)
 	else
 		enc.set_encode_id(static_cast<uint8_t>(encode_id));
 
-	cimbar::byte_istream bis(reinterpret_cast<char*>(buffer), size);
-	_fes = enc.create_fountain_encoder(bis, _compressionLevel);
+	cimbar::byte_istream bis(reinterpret_cast<const char*>(buffer), size);
+	_fes = enc.create_fountain_encoder(bis, std::string_view(filename, fnsize), _compressionLevel);
```
sz3, line 135: Historically I'd just use `string` here, but let's try out some c++17-isms

src/lib/cimbar_js/cimbar_recv_js.cpp
```diff
@@ -180,6 +182,20 @@ int cimbard_get_filesize(uint32_t id)
 	return md.file_size();
 }
 
+int cimbard_get_filename(const uchar* finbuffer, unsigned size, char* filename, unsigned fnsize)
+{
+	std::string fn = cimbar::zstd_header_check::get_filename(finbuffer, size);
+	if (!fn.empty())
+		fn = std::filesystem::path(fn).filename().string(); // strip path
```
sz3, line 189: This is the basename calculation. Might hide it in a util function, just because the syntax is so mysterious.

src/lib/compression/zstd_header_check.h
```diff
@@ -0,0 +1,35 @@
...
+
+		std::string res;
+		res.resize(500, '\0');
+		size_t sz = ZSTD_readSkippableFrame(res.data(), res.size(), nullptr, data, len);
+		if (sz <= 1)
+			return "";
+
+		switch (res[0])
+		{
+			case 1:
```
sz3, line 25: It's somewhat moot at the moment, but for future proofing we're treating the first byte as a version number. For now, it's:
`01 <filename bytes>`

We have some flexibility -- zstd has a "magic value" attached to all skippable frames, that could be used for the same purpose. But using that value locks us into having to decode it, vs a potential string match to identify the start of the frame.

src/lib/compression/zstd_header_check.h
```diff
@@ -0,0 +1,35 @@
...
+class zstd_header_check
+{
+public:
+	static std::string get_filename(const unsigned char* data, size_t len)
+	{
+		if (!ZSTD_isSkippableFrame(data, len))
+			return "";
+
+		std::string res;
+		res.resize(500, '\0');
```
sz3, line 18: *I think* this is the current limiter for max filename length.

web/recv.js
```diff
@@ -65,8 +65,14 @@ var Sink = function () {
           Recv.set_HTML("errorbox", "reassemble_file failed :( " + res);
         }
         else {
+          var name = id + "." + size;
+          const fnsize = Module._cimbard_get_filename(buff.byteOffset, buff.length, _errBuff, _errBuffSize);
+          if (fnsize > 0) {
+            const temparr = new Uint8Array(Module.HEAPU8.buffer, _errBuff, fnsize);
```
sz3, line 71: Reusing _errBuff here, since it's 1024 bytes long.

src/lib/encoder/EncoderPlus.h
```diff
@@ -43,7 +44,7 @@ inline unsigned EncoderPlus::encode(const std::string& filename, std::string out
 inline unsigned EncoderPlus::encode_fountain(const std::string& filename, const std::function<bool(const cv::Mat&, unsigned)>& on_frame, int compression_level, double redundancy)
 {
 	std::ifstream infile(filename, std::ios::binary);
-	fountain_encoder_stream::ptr fes = create_fountain_encoder(infile, compression_level);
+	fountain_encoder_stream::ptr fes = create_fountain_encoder(infile, std::filesystem::path(filename).filename().string(), compression_level);
```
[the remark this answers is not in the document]
sz3: Update: created a `basename()` function to wrap this.

PR #137, opened 2025-08-13
Bugfixes for first beta release of web decoder/receiver

sz3:
Includes:
* various improvements (hopefully) for pwa caching and updates
* honoring user theme for light/dark background (pet peeve of mine that more sites don't do this. Be the change you want to see in the world, etc etc)
* *slightly* better error reporting
    * it's still not great.
* add background/watchdog timer on ios to restart the camera feed. Currently ios only, since it interferes with ...
* ... see commit log for failed attempts to resolve an intermittent freeze on desktop chrome/firefox. It'll be a known issue for the first beta release, *hopefully* it's not common. Might be wishful thinking, but the fix is going to take some trial and error
* add version string to debug menu
* some misc code cleanup

This is probably the last PR before release.


web/recv.html
```diff
@@ -337,6 +346,7 @@
       padding: 20px;
       position: absolute;
       top: 0;
+      max-width: 95%;
```
sz3, line 349: Wanted to use 100vw, somehow that doesn't work.

web/recv.js
```diff
@@ -185,39 +182,53 @@ var Recv = function () {
...
     },
 
     download_bytes: function (buff, name) {
       var blob = new Blob([buff], { type: 'application/octet-stream' });
-      var bloburl = window.URL.createObjectURL(blob);
-      _downloadHelper(name, bloburl);
-      setTimeout(function () {
-        return window.URL.revokeObjectURL(bloburl);
-      }, 1000);
+      Zstd.download_blob(name, blob);
```
sz3, line 225: Removed code duplication here.

web/recv.js
```diff
@@ -185,39 +182,53 @@ var Recv = function () {
...
     },
 
     on_decode: function (wid, data) {
       console.log('Main thread received message from worker' + wid + ':', data);
       // if extract but no bytes, log extract counter
       if (data.nodata) {
         _recentExtract = _counter;
+        return;
```
sz3, line 233: No `return` here is a bug

PR #143, opened 2025-09-27
Config system update for proper runtime configuration

sz3:
Config params have been set at compile time, with a few piecemeal overrides to allow us to switch between mode B and the legacy mode 4C. This was never going to work out over time, so, finally:

* we have simple runtime config system. Configs are still defined at compile time (the decoder's optimizations require it to some extent), and the config functions still act the way they did before. But under the hood, we will now call `cimbar::Config::update(68)` or `cimbar::Config::update(4)` to select the config (where `mode B` == `68`, and `mode 4C` == `4`).
* the config numbers are still somewhat magical. At some point we'll add constants for them.
* the config is a *static threadlocal*, so it must be set per thread.
* while there's no UI support for it yet, the new smaller (wide) config `mode Bm` == `67`. (in a few more PRs, it will exist in the UI)
* ... we can remove quite a few hacky per-function/class param overrides, just leaning on the config to be set correctly.


CMakeLists.txt
```diff
@@ -17,7 +17,7 @@ add_definitions("-DLIBCIMBAR_PROJECT_ROOT=\"${libcimbar_SOURCE_DIR}\"")
 if("${CMAKE_BUILD_TYPE}" STREQUAL "")
 	set(CMAKE_BUILD_TYPE "RelWithDebInfo")
 	set(CMAKE_C_FLAGS "${CFLAGS} -Wall -std=c99 -Os -fPIC")
-	set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-Wall -Wextra -g -O2 -fPIC")
+	set(CMAKE_CXX_FLAGS_RELWITHDEBINFO "-Wall -Wextra -Wpedantic -Wdeprecated -g -O2 -fPIC")
```
sz3, line 20: Seems like a good idea in general

src/exe/cimbar/cimbar.cpp
```diff
@@ -156,7 +154,7 @@ int decode(const FilenameIterable& infiles, const std::function<int(cv::UMat, un
 				shouldPreprocess = true;
 		}
 
-		int bytes = decodefun(img, color_mode, shouldPreprocess, color_correct);
+		int bytes = decodefun(img, shouldPreprocess, color_correct);
```
sz3, line 157: The general pattern here is basically: if the config defines a param, just rely on the config rather than allowing other trapdoor ways to set it.

src/exe/cimbar_recv2/recv2.cpp
```diff
@@ -39,14 +39,10 @@ int main(int argc, char** argv)
 {
 	cxxopts::Options options("cimbar video decoder", "Use the camera to decode data!");
 
-	unsigned colorBits = cimbar::Config::color_bits();
-	unsigned ecc = cimbar::Config::ecc_bytes();
 	unsigned defaultFps = 15;
 	options.add_options()
 		("i,in", "Video source.", cxxopts::value<string>())
 		("o,out", "Output directory (decoding).", cxxopts::value<string>())
-		("c,colorbits", "Color bits. [0-3]", cxxopts::value<int>()->default_value(turbo::str::str(colorBits)))
```
sz3, line 48: Since cimbar_recv and recv2 do the same thing, I'm leaning towards having recv2 be the less configurable version. (e.g. less overrides for debugging, etc)

So more ad-hoc stuff like changing ecc (or colors, or whatever else) will only be supported in one, with this executable being a way to test the decoder C api functionality that the webapp does ... without the webapp.

src/lib/cimb_translator/Config.cpp
```diff
@@ -1,43 +0,0 @@
-/* This code is subject to the terms of the Mozilla Public License, v.2.0. http://mozilla.org/MPL/2.0/. */
-#include "Config.h"
-#include <cmath>
-
-namespace cimbar {
-
-unsigned Config::total_cells()
```
sz3, line 7: Because these are now per-config, and because we'd like to query some things without actually changing the config (ex: "Is the fountain chunk size different between 4C and B?"), we move all these functions down a layer into the config class.

src/lib/cimb_translator/Config.h
```diff
@@ -8,119 +8,169 @@ namespace cimbar
 	class Config
 	{
 	protected:
-		using GridConf = Conf8x8;
+		static cimbar::conf& active_conf()
+		{
+			static thread_local cimbar::conf cc = cimbar::Conf8x8();
```
sz3, line 13: arguably this should be `= temp_conf(68)`. Maybe later...

src/lib/cimb_translator/Config.h
```diff
@@ -8,119 +8,169 @@ namespace cimbar
...
 
-		static unsigned decode_window_bits();
+		static unsigned total_cells()
+		{
+			return active_conf().total_cells();
+		}
 
-		static unsigned capacity(unsigned bitspercell=0);
+		// leave bitspercell here
+		static unsigned capacity(unsigned bitspercell=0)
```
sz3, line 148: As part of the split symbol/color, 2-stage encode (+decode), we actually do vary bits per cell by how many symbol vs color bits there are. The default will calculate it from `symbol_bits + color_bits`, since we usually want the full count.

src/lib/cimb_translator/Config.h
```diff
@@ -8,119 +8,169 @@ namespace cimbar
...
+		static unsigned fountain_chunks_per_frame(unsigned bitspercell=0)
 		{
-			if (legacy_mode or !GridConf::fountain_chunks_per_frame)
-				return 10;
-			return bitspercell * GridConf::fountain_chunks_per_frame;
+			return active_conf().fountain_chunks_per_frame(bitspercell);
 		}
 
-		static unsigned fountain_chunk_size(unsigned ecc, unsigned bitspercell, bool legacy_mode);
+		static unsigned fountain_chunk_size(unsigned bitspercell=0)
```
sz3, line 168: ...and the same here as well.

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -159,19 +149,18 @@ int cimbare_configure(unsigned color_bits, unsigned ecc, int compression, bool l
 		return initRes;
 
 	// check if we need to refresh the stream
-	bool refresh = (color_bits != _colorBits or ecc != _ecc or compression != _compressionLevel or legacy_mode != _legacyMode);
+	bool refresh = (mode_val != _modeVal or compression != _compressionLevel);
```
sz3, line 152: One interesting exception I'm leaving in: we can still override compression on the encoder side.

The decoder *doesn't* care how much compression is used -- just that the data is in a zstd compressed frame. So if you wanted to, you could for example crank the compression to the maximum level (22), take the performance hit on the sender side, and possibly save a few bytes over the air.

... in practice I'm not sure how much it really would help, though.

src/lib/cimbar_js/cimbar_recv_js.cpp
```diff
@@ -29,29 +29,18 @@ namespace {
 	TimeAccumulator _tImgDecode;
 
 	// settings
-	unsigned _colorBits = 2;
 	int _modeVal = 68;
```
sz3, line 32: Trivia: the magic modeVal number originated in the android app (because I didn't want to use strings), and gradually worked its way here.

`6` => 0.6
`8` => 8x8 grid

(of course, the widescreen config is set at `67` so my perfect flawless numbering scheme has already failed...)

src/lib/encoder/Decoder.h
```diff
@@ -65,18 +57,29 @@ inline Decoder::Decoder(int ecc_bytes, int color_bits, bool interleave)
...
+	unsigned eccBlockSize = cimbar::Config::ecc_block_size();
+	unsigned colorBits = cimbar::Config::color_bits();
+	unsigned interleaveBlocks = _interleave? cimbar::Config::interleave_blocks() : 0;
+	unsigned interleavePartitions = cimbar::Config::interleave_partitions();
+
+	unsigned bitsPerSymbol = cimbar::Config::symbol_bits();
+	unsigned symCapacity = cimbar::Config::capacity(bitsPerSymbol);
+	unsigned colorCapacity = cimbar::Config::capacity(colorBits);
+	unsigned bitsPerOp = cimbar::Config::bits_per_cell();
+	unsigned fountain_chunks_per_frame = cimbar::Config::fountain_chunks_per_frame(bitsPerOp);
```
sz3, line 75: I'm not sure that we *need* to define these all at the top, old school C style. But we *are* pulling from a config object every time we do this, and it probably helps caching maybe possibly to have it all out of the way up front.

I did do some A/B performance tests, but didn't seem an appreciable difference. Still, I feel better about this way.

src/lib/encoder/Encoder.h
```diff
@@ -41,22 +40,17 @@ class Encoder
...
 	, _bitsPerSymbol(bits_per_symbol? bits_per_symbol : cimbar::Config::symbol_bits())
 	, _bitsPerColor(bits_per_color >= 0? bits_per_color : cimbar::Config::color_bits())
 	, _dark(cimbar::Config::dark())
-	, _coupled(false)
+	, _coupled(cimbar::Config::legacy_mode())
 	, _colorMode(cimbar::Config::color_mode())
 {
 }
 
-inline void Encoder::set_legacy_mode()
```
sz3, line 55: Bonus: got rid of this weird function

src/lib/encoder/test/DecoderTest.cpp
```diff
@@ -64,10 +65,12 @@ TEST_CASE( "DecoderTest/testDecode.4c", "[unit]" )
 {
 	// legacy format
 	MakeTempDirectory tempdir;
+	ConfigScope cs(4);
```
sz3, line 68: ...having a static (threadlocal in this case, but static) config means we need to lean on RAII to put things back in our tests.

As a bonus, the RAII helper can be how we set the config to begin with. Ergo: `ConfigScope`.

src/lib/encoder/test/EncoderTest.cpp
```diff
@@ -103,7 +109,10 @@ TEST_CASE( "EncoderTest/testPiecemealFountainEncoder", "[unit]" )
 {
 	// use the fountain_encoder_stream directly on a char*,int pair
 	MakeTempDirectory tempdir;
-	EncoderPlus enc(40, 4, 2);
+	ConfigScope cs;
+	cs.active_conf().ecc_bytes = 40;
```
sz3, line 113: We can still use weird ecc settings (if one insists) like this.

PR #144, opened 2025-09-28
Simplify decode c api

sz3:
For the first pass at the web decoder (https://github.com/sz3/libcimbar/pull/134), I introduced a new C api for decoding. It currently works like this:

* `cimbard_scan_extract_decode` -> on independent threads, input rgb image data, output 1-N fountain buffers
* `cimbard_fountain_decode` -> on the (1) reassembly thread, to reassemble the fountain buffers
* `cimbard_finish_copy` -> on the reassembly thread, when `cimbard_fountain_decode` reports success. output one contiguous buffer of compressed file data

And then a set of zstd apis, which frankly could be ditched for an off-the-shelf zstd implementation:
* `cimbarz_init_decompress` -> input the contiguous buffer of compressed data. Set state -- we decompress one file at a time.
* `cimbarz_decompress_read` -> read the next chunk of decompressed data

<hr>

This is changing for a few reasons:
1. it can be simpler. The api surface has *some* required "bloat" regardless: besides the basic workflow outlined above, we additionally have (necessary, non-negotiable) api functions for various buffer sizes. So anywhere we can simplify is good (imo)
2. the lifetime of the compressed file data buff is weird.
    * cimbarz_init_decompress and cimbarz_decompress_read don't copy the data, they just read from the buff. That means if something unexpected happens to the buffer, we're gonna crash. Or more to the point, it means that the caller (the user of the api) now needs to worry about the lifetime of this intermediate object.

Luckily, there's no real need to expose this intermediate compressed representation (short of for testing purposes -- but we can work around that). So we can collapse a few methods down. The new api is:

* `cimbard_scan_extract_decode` -> on independent threads, input rgb image data, output 1-N fountain buffers. _(same as before)_
* `cimbard_fountain_decode` -> on the (1) reassembly thread, to reassemble the fountain buffers _(same as before)_
* `cimbard_decompress_read` -> on the reassembly thread, when `cimbard_fountain_decode` reports success.  read the next chunk of decompressed data, or the first chunk (if it's the first call)

So: we've collapsed the last 3 calls into 1, and gone from 5 to 3 as far as the main flow is concerned. Also, the intermediate buffer is now internal state, so the caller doesn't have to think about it too much.


src/exe/cimbar_recv2/recv2.cpp
```diff
@@ -150,17 +150,28 @@ int main(int argc, char** argv)
...
 			std::vector<unsigned char> data;
-			data.resize(size);
-			int res = cimbard_finish_copy(fileId, data.data(), size);
-			if (res != 0)
-				std::cerr << "failed fountain_finish_copy " << res << std::endl;
+			data.resize(cimbard_get_decompress_bufsize());
 
-			decompress_on_store<std::ofstream>(outpath, true)(filename, data);
+			int res = 1;
+			while ((res = cimbard_decompress_read(fileId, data.data(), data.size())) > 0)
```
sz3, line 171: Actually using the full api here now.

src/lib/cimbar_js/cimbar_recv_js.cpp
```diff
@@ -215,4 +287,12 @@ int cimbard_configure_decode(int mode_val)
 	return 0;
 }
 
+// testing
+unsigned char* cimbard_get_reassembled_file_buff()
+{
+	if (_reassembled.empty())
+		return nullptr;
+	return _reassembled.data();
```
sz3, line 295: Testing only, please

src/lib/cimbar_js/cimbar_recv_js.cpp
```diff
@@ -186,15 +244,29 @@ int cimbard_get_filename(const uchar* finbuffer, unsigned size, char* filename,
...
-// bouth fountain_*() calls should be from the same js webworker/thread
-int cimbard_finish_copy(uint32_t id, uchar* buffer, unsigned size)
+int cimbard_decompress_read(uint32_t id, unsigned char* buffer, unsigned size)
 {
-	if (!_sink)
-		return -1;
-	if (!_sink->recover(id, buffer, size))
-		return -2;
-	return 0;
+	int	res = recover_contents(id);
```
sz3, line 249: Notably both decompress_read() *and* get_filename() will call `recover_contents()` (result will be cached), since they both need to pull the data from the zstd stream. Definitionally, then, you can only call either after `cimbard_fountain_decode()` has finished reassembling the file.

web/main.js
```diff
@@ -366,3 +363,7 @@ window.addEventListener("drop", function (e) {
   Main.dragDrop(e);
   document.body.style["opacity"] = 1.0;
 }, false);
+
+window.addEventListener('resize', () => {
+  Main.resize();
```
sz3, line 368: Also: listen for the resize event

web/test/test_base.js
```diff
@@ -1,19 +1,13 @@
 QUnit.module("base");
 QUnit.config.reorder = false;
+QUnit.config.testTimeout = 10000;
 
-var Zstd = function () {
```
sz3, line 4: Another upside: we're now testing more of the zstd code path in the JS test.

web/recv.js
```diff
@@ -55,31 +55,22 @@ var Sink = function () {
...
+          const temparr = new Uint8Array(Module.HEAPU8.buffer, _errBuff, fnsize);
+          name = new TextDecoder("utf-8").decode(temparr);
         }
+        Zstd.decompress(name, id);
       } catch (error) {
         console.log("failed finish copy or download?? " + error);
       }
-      // this needs to happen after decompress() completes
-      // currently decompress is sync, so it's fine. But...
-      Module._free(dataPtr);
```
sz3, line 82: ^

Happy to make this memory cleanup race go away.

Issue #141, opened 2025-08-28
Some feature requests for low-end devices


llllllllqq:
This program is an amazing creation, it can help me to transfer log files from an old Windows7 x86 device to my phone without internet.
In my situation, I have some other ideas for low-end devices.

1. Provide a pure dynamic standard QR code mode for very small files (e.g. <100KB) or on user request. Just like [qrs](https://github.com/qifi-dev/qrs) or [txqr](https://github.com/divan/txqr) . This can be more usable and robust with bad screen resolution or with bad performance camera.
2. Provide a single html file (just like what `cimbar_js.html` does) cimbar `cli` that can export `png` for download. This can be useful on old browsers/CPUs/GPUs that not support WebGL well, and is more easy to use for devices mainly on browser without keyboard.
3. Auto split function (like part001 part002... which can be merged manually) to prevent big file OOM. Now I force my system log size under 100KB but it make the system slow.

sz3:
> 1. Provide a pure dynamic standard QR code mode for very small files (e.g. <100KB) or on user request. Just like [qrs](https://github.com/qifi-dev/qrs) or [txqr](https://github.com/divan/txqr) . This can be more usable and robust with bad screen resolution or with bad performance camera.

I think I probably wouldn't include a QR code -- I'll likely let other projects do that. But I am working on a smaller format that works on worse cameras/screens.

There are also some robustness improvements for the decoder that I'd like to make (not sure when that'll happen though)

> 2. Provide a single html file (just like what cimbar_js.html does) cimbar cli that can export png for download. This can be useful on old browsers/CPUs/GPUs that not support WebGL well, and is more easy to use for devices mainly on browser without keyboard.

This is something I hadn't thought about, but it's an interesting idea! I do worry that a version of the web decoder will be difficult/impossible (the apis we need probably don't exist on those old devices), but the *encoder* should be able to work.

> 3. Auto split function (like part001 part002... which can be merged manually) to prevent big file OOM. Now I force my system log size under 100KB but it make the system slow.

🤔 

I'm working on something that will hopefully help with OOMs (even without any splitting) -- I don't know if it'll help a little or a lot yet though.

sz3:
A new release is out (https://github.com/sz3/libcimbar/releases/tag/v0.6.3), and it may help with OOM errors.

It also includes the experimental format `mode Bm` that should work more reliably (at cost of slower max performance)

xPeiPeix:
[...]

Issue #132, opened 2025-07-12
你好开发者，为什么我用html发送19MB的rar格式，多次尝试一直无反应，请问问题出在哪里？？


caoxino:
你好开发者，为什么我用html发送19MB的rar格式，选择19MB文件后，浏览器页面卡一下，然后再无动静，后续多次尝试一直无反应，请问问题出在哪里？感谢你的项目，后续可否增加exe版本？

sz3:
> Hello developers, why I sent a 19MB rar file in HTML format but it didn't work after multiple attempts. What's the problem? 

> Hello developer, why when I use HTML to send a 19MB rar format, after selecting the 19MB file, the browser page freezes for a while, and then nothing happens. After multiple attempts, there is still no response. What is the problem? Thank you for your project. Can you add an exe version in the future?

I probably need to make sure error to shows on the page and instead of being hidden in the console...

For files in the ~20 MB range or larger, the browser is inconsistent about giving us a contiguous chunk of memory -- sometimes it works (after taking a few seconds), but it also can fail (what you're seeing, I think) with an out of memory (OOM) error.

It's pretty random in my experience, so you could try a different browser to see if that helps. Otherwise, the simplest thing is probably to split the file, send it in smaller chunks, and put it back together on the receiver side.

> Can you add an exe version in the future?

Perhaps. I've been hesitant to make a pure native version (since I'd have to dev/test on an additional platform), but something like a webview wrapper is probably reasonable.

Issue #118, opened 2025-01-27
Fail to encode file in web app


freejool:
It's ok to encode a 2MB file, but when uploading a 30MB file -->
<img width="471" alt="Image" src="https://github.com/user-attachments/assets/c87722fb-831f-4f6b-b4a9-74dadb135c78" />

sz3:
Unfortunately we need a block of contiguous memory, and sometimes the OS/browser doesn't want to do that (sometimes?) when the size is large enough.

Not sure what a good solution is (it'll become more important if/when I'm able to increase the transfer speed)

freejool:
Figuring out how much is the usual max size and how to increase that restriction may be helpful.

freejool:
sorry for the delay ;) 

In my case(Safari 26.0.1), the file size limit is still approximately 20MB. Tested with https://github.com/sz3/libcimbar/releases/download/v0.6.3/cimbar_js.html and https://cimbar.org/

<img width="1439" height="473" alt="Image" src="https://github.com/user-attachments/assets/c8ea50a7-240c-433d-a7d0-72071c653984" />

Issue #96, opened 2024-05-19
 Aborted(OOM) error


Katzenwerfer:
Trying to a ~90 MB file and I'm getting this error.
More specifically these 2.
```
17:13:38.013 file input: [object HTMLInputElement] main.js:126:13
17:13:39.065 Aborted(OOM) cimbar_js.js:1:5472
    abort https://cimbar.org/cimbar_js.js:1
    abortOnCannotGrowMemory https://cimbar.org/cimbar_js.js:1
    _emscripten_resize_heap https://cimbar.org/cimbar_js.js:1
    <anonymous> https://cimbar.org/cimbar_js.wasm:64936
    <anonymous> https://cimbar.org/cimbar_js.wasm:12358
    onload https://cimbar.org/main.js:25
    (Async: EventHandlerNonNull)
    importFile https://cimbar.org/main.js:22
    fileInput https://cimbar.org/main.js:129
    onchange https://cimbar.org/:1
17:13:39.067 Uncaught RuntimeError: Aborted(OOM). Build with -sASSERTIONS for more info. cimbar_js.js:1:5558
    abort https://cimbar.org/cimbar_js.js:1
    abortOnCannotGrowMemory https://cimbar.org/cimbar_js.js:1
    _emscripten_resize_heap https://cimbar.org/cimbar_js.js:1
    onload https://cimbar.org/main.js:25
    importFile https://cimbar.org/main.js:22
    fileInput https://cimbar.org/main.js:129
    onchange https://cimbar.org/:1
```

sz3:
Neat!

cimbar currently can't support files larger than 33MB post-compression (specifically, the number is `33554431` bytes, 2^25-1), so unless you 90MB file compresses *very* well it's probably too large even if we don't hit the OOM error. But it'd be nice to not run out of memory. :thinking:

PR #145, opened 2025-09-28
Improve encode c api

sz3:
The flip side of https://github.com/sz3/libcimbar/pull/144: *complex-ify* the encoder. :melting_face: 

We (users of the cimbar.js encoder, including me) have run into situations where the browser hits an out of memory (OOM) error on large input files. Example issues:
https://github.com/sz3/libcimbar/issues/141
https://github.com/sz3/libcimbar/issues/132
https://github.com/sz3/libcimbar/issues/118
https://github.com/sz3/libcimbar/issues/96

While I can't solve all OOM issues -- wirehair needs the compressed representation stored in contiguous memory -- writing the decoder (#134) reminded/convinced me that we can do better when loading the *uncompressed* data for encoding.

So:
* we're now loading input files in chunks
* this requires us to change the c api for encoding
* as a result, OOM errors in the browser should be less frequent. I won't say solved; but, better. 

The new API surface:
* `cimbare_init_encode(filename, encode_id)` -> starts an encode, waits for data
* `cimbare_encode(data)` -> provides data. When the size of data provided is == `cimbare_encode_bufsize()`, we treat this as an incremental write and don't finalize anything (and return `1`). Otherwise, we consider the write complete.
* `cimbare_encode(nullptr, 0)` therefore functions as a sort of "flush()" operation if you want to make sure the data is ready.
* when cimbare_encode() returns `0`, we can start doing our `cimbare_next_frame()` calls

### DIFF
--- a/src/lib/cimbar_js/cimbar_js.cpp
+++ b/src/lib/cimbar_js/cimbar_js.cpp
@@ -140,6 +140,8 @@ int cimbare_init_encode(const char* filename, unsigned fnsize, int encode_id)
 
 	if (fnsize > 0 and filename != nullptr)
 		_comp->write_header(filename, fnsize);
+
+	_fes.reset();
 	return 0;
 }
 
--- a/src/lib/cimbar_js/cimbar_js.cpp
+++ b/src/lib/cimbar_js/cimbar_js.cpp
@@ -2,6 +2,7 @@
 #include "cimbar_js.h"
 
 #include "cimb_translator/Config.h"
+#include "compression/zstd_compressor.h"
 #include "encoder/Encoder.h"
 #include "gui/window_glfw.h"
 #include "util/byte_istream.h"
@@ -15,7 +16,12 @@ namespace {
 	std::shared_ptr<fountain_encoder_stream> _fes;
 	std::optional<cv::Mat> _next;
 
+	// compressing the file
+	std::unique_ptr<cimbar::zstd_compressor<std::stringstream>> _comp;
+
 	int _frameCount = 0;
+	// start encode_id is 109. This is mostly unimportant (it only needs to wrap between [0,127]), but useful
+	// for the decoder -- because it gives it a better distribution of colors in the first frame header it sees.
 	uint8_t _encodeId = 109;
 
 	// settings, will be overriden by first call to configure()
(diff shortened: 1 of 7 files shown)


src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -117,17 +127,51 @@ int cimbare_encode(const unsigned char* buffer, unsigned size, const char* filen
 		return -5;
 	}
 
-	Encoder enc;
 	if (encode_id < 0)
-		enc.set_encode_id(++_encodeId); // increment _encodeId every time we change files
+		++_encodeId; // increment _encodeId every time we change files
 	else
-		enc.set_encode_id(static_cast<uint8_t>(encode_id));
+		_encodeId = encode_id;
```
sz3, line 133: This might be the strangest side effect of this change -- `_encodeId` will now persist if you override it, so you can change the auto-increment starting point. You probably shouldn't, though. The right behavior will generally be to use `-1` (for auto-increment).

(should probably define a constant for that)

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -117,17 +127,51 @@ int cimbare_encode(const unsigned char* buffer, unsigned size, const char* filen
...
+
+	// otherwise, we're ready
+	unsigned fountainChunkSize = cimbar::Config::fountain_chunk_size();
+	size_t compressedSize = _comp->size();
+	if (compressedSize < fountainChunkSize)
+		_comp->pad(fountainChunkSize - compressedSize + 1);
+
+	// create the encoder stream
+	_fes = fountain_encoder_stream::create(*_comp, fountainChunkSize, _encodeId);
+	_comp.reset();
```
sz3, line 172: After we return 0 from `cimbare_encode`, the next call to it will return `-1` (as if we never called `cimbare_init_encode`)

src/lib/cimbar_js/test/cimbar_jsTest.cpp
```diff
@@ -65,11 +66,30 @@ TEST_CASE( "cimbar_jsTest/testRoundtrip", "[unit]" )
...
+	std::string contents = random_string(size);
+	std::string filename = "/tmp/foobar-c语言版.txt";
+	assertEquals( 0, cimbare_init_encode(filename.data(), filename.size(), 100) );
+	assertEquals( 1, cimbare_encode(reinterpret_cast<unsigned char*>(contents.data()), contents.size()) );
+	assertEquals( -1, cimbare_next_frame() );
+
+	assertEquals( 0, cimbare_encode(nullptr, 0) );
+	assertEquals( 1, cimbare_next_frame() );
+
+	assertEquals( -1, cimbare_encode(nullptr, 0) );
```
sz3, line 93: Demonstrating the corner case.

PR #148, opened 2025-11-21
Autodetect mode in the web decoder + fixes

sz3:
Autodetect will work close to how it does in the CFC android app. I debated leaving explicit mode selection buttons in the UI, but (for now?) collapsed it down to a binary autodetect/manual selection to save space and menu cruft.

Bugfixes:
* limit the number of inflight/stored frames during processing. This mirrors what CFC does, but more crucially: the webapp would crash after a while if it couldn't keep up, since it would allocate more and more memory for frames.
    * the other obnoxious thing (if the crashing wasn't bad enough) is that the webapp's feedback (the preview frame, the progress bars, and the onscreen colors) would get out of sync with what it was actually processing. So e.g. you'd see the progress go up even after you put your phone down, because it had stored a bunch of frames and was only now getting around to processing them.
    * ... anyway, limiting the size of queues is important
* if multiple transfers are in progress at the same time, we now actually draw more than one progress bar. :upside_down_face: ... this is another silly one.
* if the fountain buffer size (in the decoder JS) is too small, allocate a bigger one.

Misc:
* abandon gcc9 build in GH actions, it's old.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,9 +10,9 @@ jobs:
       fail-fast: true
       matrix:
         config:
-        - name: "linux gcc9"
+        - name: "linux gcc11"
           os: ubuntu-22.04
-          env: CXX="g++-9" CC="gcc-9"
+          env: CXX="g++-11" CC="gcc-11"
 
         - name: "linux gcc-latest"
           os: ubuntu-latest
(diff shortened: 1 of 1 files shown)


web/recv.js
```diff
@@ -376,38 +413,45 @@ var Recv = function () {
...
-      Module._cimbard_configure_decode(modeVal);
-      for (let i = 0; i < _workers.length; i++) {
-        // cal config decode within the workers as well
-        _workers[i].postMessage({ config: true, mode_val: modeVal });
+
+      // configure wasm in main thread
+      _mode = modeVal;
+      if (_mode > 0) {
+        Module._cimbard_configure_decode(_mode);
+        Sink.allocate();
```
sz3: ...and here's where we need to make sure our buffers look good.

PR #152, opened 2026-01-08
Alternate color scheme for encoder

sz3:
In certain circumstances, it is useful to slightly alter the color palette used for encoding vs decoding to even out the brightness of the resulting image.

This is toggled off by default (and requires manually changing a variable), but has proved very useful for testing. It's only applicable (for now) for 4-color configurations (e.g. B, 4C).

Misc other thoughts:
* this *really* shows up as useful for the test 5x5 config. Unfortunately, 5x5 cimbar has other issues and *probably* won't be a default config (though never say never), but this trick is essential.
* 8x8 results are more inconclusive.
* 6x6 is an unknown...

### DIFF
--- a/src/lib/fountain/fountain_decoder_sink.h
+++ b/src/lib/fountain/fountain_decoder_sink.h
@@ -159,10 +159,10 @@ public:
 
 		// when you provide a write callback,
 		// store() will call mark_done() afterwards
-		// -- and the assembled file will we dropped from RAM.
+		// -- and the assembled file will be dropped from RAM.
 		// but if no callback is provided, you can do something else.
 		store(md, s);
-		return (int64_t)md.id();
+		return (int64_t)0 + md.id();
 	}
 
 	bool write(const char* data, unsigned length)
--- a/web/main.js
+++ b/web/main.js
@@ -2,7 +2,7 @@ var Main = function () {
 
   // configurable
   var _interval = 66;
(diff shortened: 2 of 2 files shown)


src/lib/cimb_translator/Common.cpp
```diff
@@ -25,7 +25,19 @@ namespace {
...
+	}
+
+	RGB getColor4_enc(unsigned index)
+	{
+		// opencv uses BGR, but we don't have to conform to its tyranny
+		static constexpr array<RGB, 4> colors = {
+			RGB(0, 0xFF, 0),
+			RGB(0, 0xFF, 0xFF),
+			RGB(0xFF, 0xFF, 0),
+			RGB(0xFF, 0x55, 0xFF),
```
sz3: I say "alternate color scheme", but it's just one color.  We add some G to the magenta to make it brighter.

src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -104,7 +104,7 @@ int cimbare_render()
 	return 0;
 }
 
-int cimbare_next_frame()
+int cimbare_next_frame(bool color_balance)
```
sz3: It might be better to move this into the config step somehow, to keep it out of the method signature here. For now it seems ok...

PR #154, opened 2026-01-08
Add padding option to extract/deskew step

sz3:
This is an attempt to deal with more distortion, and I'm not entirely sure how effective it'll be. But it seems *probably* useful in my tests, and we can always remove it later. The gist is:

* during decode, we track how far off from our perfect grid we are. The class responsible for this bookkeeping is called "CellDrift"
* we have a max drift, which is determined by the size of our image. It's currently set to `7`, 1 less than the padding we have around our image.
* as such, if we have more than 7px (post extraction) of distortion, we have to give up on those cells.

BUT:
* we can pad our extract -- instead of extracting a 1024x1024 image, we can make add 4px of padding and make it e.g. 1032x1032 (with the 1024x1024 we care about at the center). This allows us to "drift" more, and theoretically (maybe) deal with more distortion.

... I'm still testing this in the android decoder, so the default is still "0" padding, or in other words no change from the existing code. But this is a good checkpoint.

### DIFF
--- a/src/lib/extractor/Deskewer.h
+++ b/src/lib/extractor/Deskewer.h
@@ -11,27 +11,28 @@
 class Deskewer
 {
 public:
-	Deskewer(cimbar::vec_xy image_size={}, unsigned anchor_size=0);
+	Deskewer(unsigned padding=0, cimbar::vec_xy image_size={}, unsigned anchor_size=0);
 
 	template <typename MAT>
-	MAT deskew(const MAT& img, const Corners& corners, unsigned padding=0);
+	MAT deskew(const MAT& img, const Corners& corners);
 
 protected:
 	cimbar::vec_xy _imageSize;
 	unsigned _anchorSize;
+	unsigned _padding;
 };
 
 template <typename MAT>
-inline MAT Deskewer::deskew(const MAT& img, const Corners& corners, unsigned padding)
+inline MAT Deskewer::deskew(const MAT& img, const Corners& corners)
 {
 	std::vector<cv::Point2f> outputPoints;
-	outputPoints.push_back(cv::Point2f(_anchorSize+padding, _anchorSize+padding));
-	outputPoints.push_back(cv::Point2f(_imageSize.width() - _anchorSize+padding, _anchorSize+padding));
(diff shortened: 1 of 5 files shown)


src/lib/cimb_translator/CimbReader.cpp
```diff
@@ -40,8 +40,7 @@ namespace {
 		}
 		cv::adaptiveThreshold(symbols, symbols, 255, cv::ADAPTIVE_THRESH_MEAN_C, cv::THRESH_BINARY, blockSize, 0);
 
-		// TODO: this should be on capacity, not image_size... right??
-		bitbuffer bb(Config::image_size_x() * Config::image_size_y() / 8);
+		bitbuffer bb(symbols.rows * symbols.cols / 8);
```
sz3: > TODO: this should be on capacity, not image_size... right??

WRONG! The code was right, the TODO was wrong. So it goes...

src/lib/cimb_translator/CimbReader.cpp
```diff
@@ -110,10 +109,11 @@ CimbReader::CimbReader(const cv::Mat& img, CimbDecoder& decoder, unsigned color_
 	, _fountainColorHeader(0U)
 	, _radioactiveBlockId(0) // can only compute once we know the file size
 	, _cellSize(Config::cell_size() + 2)
+	, _gridPadding(std::min(_image.cols - Config::image_size_x(), _image.rows - Config::image_size_y())/2)
```
sz3: One nice(?) thing about this change is that we have the CimbReader (concerned with post-extract images) intelligently figure out the padding. Which in turn means it can now accept non-exact image sizes -- although the image still needs to be centered to work.

src/lib/extractor/Deskewer.h
```diff
@@ -11,26 +11,28 @@
 class Deskewer
 {
 public:
-	Deskewer(cimbar::vec_xy image_size={}, unsigned anchor_size=0);
+	Deskewer(unsigned padding=0, cimbar::vec_xy image_size={}, unsigned anchor_size=0);
```
sz3: In usual circumstances you won't want to specify the image or anchor sizes here. So normal usage will be:

`Deskewer ds;`

or

`Deskewer ds(4);` (or whatever padding is desired)

Issue #147, opened 2025-10-27
files from phone to laptop


Mcofir:
trying to move files from android phone to laptop, but unfortunately the laptop camera cant catch the barcode shapes due to laptop camera is not so good and phone screen is small, so this combination is really bad.

i thought maybe if you can make a mode with larger shapes & more frames/smaller file sizes support , for such cases , cause this way i couldnt transfer any files.

thank you , i love the idea of the project

Anonymous3-a:
Even with mode Bm? Bm is already a bit slow compared to others...

How large are your files?

Mcofir:
> Even with mode Bm? Bm is already a bit slow compared to others...
> 
> How large are your files?

hey , yes i tried all the modes , when i put the phone infront of the camera it catches just some noisy colors,
files are under 600kb - (images)

sz3:
> it seems like manual focus is not able to switch to manual mode

This is what I'm afraid of where manual focus is concerned -- I'm not sure how well the browser APIs (and cameras!) actually work. So we may be stuck with trying to trick autofocus into doing the right thing.

I wonder if lowering the framerate might help? Obviously that slows down the transfer speed, but right now you're getting 0 bytes at all 🤔 (see https://github.com/sz3/libcimbar/issues/105 -- this would require you to download the code and host it locally though)

One thing you can try is to capture a camera frame (e.g. with a screenshot on the laptop) and we can look at it to see if a human could decode it.

Mcofir:
i dont think the issue is the frames, my camera is 60fps, even if i put my finger on the bottom of the screen making the image stop , the image is not clear enough , thats why at first i was suggesting bigger symbols method , here ill add an image while the barcode is frozen and i took a screenshot ,the symbols are not clear enough
https://files.catbox.moe/avj1yt.png

Mcofir:
im really appreciate your work and that you still attempt to fix this , thank you ! , ill be wait for updates

Issue #142, opened 2025-09-21
使用电脑摄像头在cimbar.org网站发送和接收文件不成功


tomchen008:
我使用cimbar.org网站发送和接收文件，当使用androd app时能够成功接收文件，但使用电脑摄像头时（re.cimbar.org）不成功，实时预览画面的框一直是黄色的，是否对电脑摄像头硬件上有什么要求么？还是其它原因？

sz3:
> I'm using cimbar.org to send and receive files. I can successfully receive files using the Android app, but not when using my computer camera (re.cimbar.org). The live preview frame remains yellow. Are there any hardware requirements for my computer camera? Or is there another reason?

The most common failure modes I know for webcams are:
* low camera resolution. We really need a >900p image. Though I'm working on a smaller version of the format that should help with 720p cameras.
* bad autofocus. Consumer webcams seem to struggle more than cell phone cameras when trying to focus on the screen.

Unfortunately neither of these are satisfying answers. Manual focus would be nice to have in the webapp, but I'm not sure how soon I'll be able to add it.

Putting the bad news aside:
The smaller format ("mode Bm") will be ready soon. It works a lot more consistently on my webcam...

PR #155, opened 2026-01-18
Add mode "Bu"

sz3:
A sequel of sorts to https://github.com/sz3/libcimbar/pull/146 .

Mode `B` is 1024x1024 (9300 bytes)
Mode `Bm` is 1024x720 (6444 bytes)

This new mode, `Bu` is 736x637 (4032 bytes). The emphasis is on compatibility -- the cost of performance of e.g. mode B is that on some devices (perhaps: laptop webcams "reading" from a cimbar encoder on a phone screen) will not work. By shrinking the grid, the chances of success should be much higher. There are also some scenarios where mode Bm *sort of* works, but requires a lot of concentration to keep the camera and screen perfectly aligned.

In any case, if mode B works in your setup, you should keep using it. But if it won't work, ~43 kb/s is better than 0 kb/s. :slightly_smiling_face: 

https://github.com/sz3/libcimbar/issues/147
https://github.com/sz3/libcimbar/issues/142
https://github.com/sz3/libcimbar/issues/141
(and others)

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -95,12 +95,14 @@
     /* options */
     #nav-container a.mode-b,
     #nav-container a.mode-bm,
+    #nav-container a.mode-bu,
     #nav-container a.mode-4c {
       color: #555;
     }
 
     #nav-container.mode-b a.mode-b,
     #nav-container.mode-bm a.mode-bm,
+    #nav-container.mode-bu a.mode-bu,
     #nav-container.mode-4c a.mode-4c {
       color: #9fa6b2;
     }
@@ -354,6 +356,7 @@
           <h4>Mode</h4>
           <a class="mode-b" href="javascript:void(0)" onclick="Main.setMode('B')">B</a>
           <a class="mode-bm" href="javascript:void(0)" onclick="Main.setMode('Bm')">Bm</a>
+          <a class="mode-bu" href="javascript:void(0)" onclick="Main.setMode('Bu')">Bu</a>
           <a class="mode-4c" href="javascript:void(0)" onclick="Main.setMode('4C')">4C</a>
         </li>
         <li>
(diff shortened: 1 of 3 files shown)


src/lib/cimb_translator/Config.h
```diff
@@ -101,6 +105,7 @@ namespace cimbar
 
 		static constexpr unsigned cell_size()
 		{
+			// TODO: runtime
```
sz3: I spent quite a bit of time testing 5x5 configs before settling on a small 8x8. Unfortunately I don't think it's going to work well enough :disappointed: 

There is maybe room for a large 5x5 config for "best case scenario" speed purposes (130+ kb/s is possible) . But it'll always be (afaik) much less reliable/useful compared to mode B.

src/lib/cimb_translator/Config.h
```diff
@@ -25,12 +25,16 @@ namespace cimbar
...
+					cc.fountain_chunks_scalar = -10;
 					return cc;
 				case 8:
 					cc = cimbar::Conf8x8();
 					cc.color_bits = 3;
 					cc.legacy_mode = true;
+					cc.fountain_chunks_scalar = -10;
 					return cc;
+				case 66:
+					return cimbar::Conf8x8_micro();
```
sz3: This is pretty much all you need to add a new config at this point. The rest of the PR is mostly boilerplate.

src/exe/cimbar_recv2/recv2.cpp
```diff
@@ -67,6 +67,8 @@ int main(int argc, char** argv)
 		string mode = result["mode"].as<string>();
 		if (mode == "4c" or mode == "4C")
 			config_mode = 4;
+		else if (mode == "Bu" or mode == "BU")
+			config_mode = 66;
```
sz3: Might factor this into a helper function so these cli tools don't have to duplicate all this.

PR #156, opened 2026-01-18
web encoder wakelock

sz3:
When using a phone to encode large files, it's annoying to have to wake up the screen.

### DIFF
--- a/web/main.js
+++ b/web/main.js
@@ -75,10 +75,9 @@ var Main = function () {
 
   // public interface
   return {
-    init: async function (canvas) {
+    init: function (canvas) {
       Main.setMode('B');
       Main.check_GL_enabled(canvas);
-      Main.prevent_sleep();
     },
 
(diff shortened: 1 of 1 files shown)


web/main.js
```diff
@@ -165,12 +167,28 @@ var Main = function () {
...
+        return;
+      }
+      const requestWakeLock = async () => {
+        try {
+          _wakeLock = await navigator.wakeLock.request('screen');
+          console.log('got wake lock!');
+          _wakeLock.addEventListener('release', () => {
+            _wakeLock = undefined;
+          });
+        } catch (err) { }
```
sz3: A bit clunky, but we're failing silently here partly because sometimes the reason we won't get a wake lock is that we shouldn't be trying. e.g. if the window is out of focus on desktop.

I *think* this is probably fine. If we *do* fail to get the wake lock (and don't know about it) it's not the end of the world. Not every error is a show stopper...

Issue #128, opened 2025-06-18
Bigger `cimbar_send` window


Anonymous3-a:
My screen has a fairly high resolution now (2560 * 1600), but the size of the screen does not scale to match, as I have a laptop. This means that basically everything tends to be smaller than preferred, but I have adjusted to that. Unfortunately, my phone is not able to do that. It can only get so close to the screen before the cimbar goes out of frame, which gives me two options.

1) I can try to get the exact right position, which is very hard to get with my shaky hands.
2) I can use the `cimbar` program in encoding mode, then recombine the images into a gif which I play at full resolution.

Option one is very hard to get working, and option two is very slow. Right now, I am using the https://cimbar.org/ encoder, which plays at full resolution, but if there was a way to scale the `cimbar_send` window, that would be great. A quick `$ cimbar_send ~/Pictures/Screenshot` is faster than starting up my browser and loading a website with my slow internet.

sz3:
Good news: I'm currently working on this!

Bad news: The version I have working is the silly version (just scale to window size, but don't lock the aspect ratio). It's also wrapped up in a bigger changeset (https://github.com/sz3/libcimbar/tree/grid-flexibility-for-5x6-or-8x9).

... but you might be able to try out that branch. I'm also hoping to get things cleaned up and merged soon.

Anonymous3-a:
It works amazing! Thank you.

Anonymous3-a:
I'd like to note that fullscreening in Wayland still doesn't have the aspect lock, you'd need to pad the sides. Normal resizing works fine, so I can just move the window to a corner and resize.

bonswouar:
> That looks like high DPI silliness. Does this branch work? https://github.com/sz3/libcimbar/tree/high-dpi-fix

Yes it totally works, now the frame uses the whole window, great!

PR #160, opened 2026-04-02
cimbar_send cli: fix high DPI scaling on wayland

sz3:
* switch to glfwSetFramebufferSizeCallback(), but only on desktop. wasm code relies on browser-side resize logic)
* using glfwWindowHint() on desktop-only code paths (also not in the wasm)
* not sure about future impacts on windows/mac, let me know if you try to get it working and it doesn't look right

See also: #128, #159

### DIFF
--- a/src/lib/gui/window_glfw.h
+++ b/src/lib/gui/window_glfw.h
@@ -102,7 +102,7 @@ public:
 			glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 			glViewport(xOffset, yOffset, viewportWidth, viewportHeight);
 		};
-		glfwSetWindowSizeCallback(_w, fun);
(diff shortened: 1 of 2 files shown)


src/lib/gui/window_glfw.h
```diff
@@ -102,7 +102,7 @@ class window_glfw
 			glClear(GL_COLOR_BUFFER_BIT | GL_DEPTH_BUFFER_BIT);
 			glViewport(xOffset, yOffset, viewportWidth, viewportHeight);
 		};
-		glfwSetWindowSizeCallback(_w, fun);
+		glfwSetFramebufferSizeCallback(_w, fun);
```
sz3: This OTOH seems to be necessary for wayland, and doesn't *seem* to hurt anything in Xorg.

PR #161, opened 2026-04-25
Minor webui improvements

sz3:
2 things:

1. for the web decoder, the "crosshairs" now move to the edge of the camera frame (on the smallest dimension), while also matching the aspect ratio of the detected mode. (e.g. B == 1:1, Bm == 1.4)
   * this means you can rely on them (somewhat) for positioning the barcode inside the frame. It's still not perfect, but it's better.
2. fixed a bug in the encoder and decoder around buffer reuse. Caching a `Uint8Array(Module.HEAPU8.buffer, ...)` is apparently malpractice (the browser might move the buffer?), and this probably led to a nonzero amount of mysterious crashes in the past (and future: until the next release :upside_down_face: ). The fix (or: *a* fix...) is to repair the cached buffer object on reuse.

### DIFF
--- a/web/main.js
+++ b/web/main.js
@@ -25,12 +25,21 @@ var Main = function () {
     }
   }
 
-  function importFile(file) {
-    let chunkSize = Module._cimbare_encode_bufsize();
+  function compress_buff(chunkSize) {
     if (_compressBuff === undefined) {
       const dataPtr = Module._malloc(chunkSize);
       _compressBuff = new Uint8Array(Module.HEAPU8.buffer, dataPtr, chunkSize);
     }
+    else if (_compressBuff.buffer !== Module.HEAPU8.buffer) {
+      _compressBuff = new Uint8Array(Module.HEAPU8.buffer, _compressBuff.byteOffset, _compressBuff.byteLength);
+    }
+    return _compressBuff;
+  }
+
+  function importFile(file) {
+    let chunkSize = Module._cimbare_encode_bufsize();
+    let compBuff = compress_buff(chunkSize);
+
     let offset = 0;
     let reader = new FileReader();
 
@@ -41,8 +50,9 @@ var Main = function () {
       if (datalen > 0) {
         // copy to wasm buff and write
         const uint8View = new Uint8Array(event.target.result);
-        _compressBuff.set(uint8View);
-        const buffView = new Uint8Array(Module.HEAPU8.buffer, _compressBuff.byteOffset, datalen);
+        compBuff = compress_buff(chunkSize);
+        compBuff.set(uint8View);
+        const buffView = new Uint8Array(Module.HEAPU8.buffer, compBuff.byteOffset, datalen);
         Main.encode_bytes(buffView);
 
         offset += chunkSize;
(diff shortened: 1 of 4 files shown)


web/main.js
```diff
@@ -25,12 +25,21 @@ var Main = function () {
...
   }
 
-  function importFile(file) {
-    let chunkSize = Module._cimbare_encode_bufsize();
+  function compress_buff(chunkSize) {
     if (_compressBuff === undefined) {
       const dataPtr = Module._malloc(chunkSize);
       _compressBuff = new Uint8Array(Module.HEAPU8.buffer, dataPtr, chunkSize);
     }
+    else if (_compressBuff.buffer !== Module.HEAPU8.buffer) {
```
sz3, line 33: The thought that comes to mind is, "maybe we should just store a `{byteOffset, length}` pair instead of  the Uint8Array(), which we can't trust anyway. But practically speaking, we're going to end up with some boilerplate like this function no matter what.

web/recv-worker.js
```diff
@@ -99,11 +99,21 @@ var RecvWorker = function () {
     },
 
     imgBuff: function () {
-      return _buffs['img'];
+      let buff = _buffs['img'];
+      if (buff.buffer !== Module.HEAPU8.buffer) {
+        _buffs['img'] = new Uint8Array(Module.HEAPU8.buffer, buff.byteOffset, buff.byteLength);
+        buff = _buffs['img'];
```
sz3, line 105: this, OTOH, is pretty gross, and makes me question the pattern.

web/recv.js
```diff
@@ -119,6 +127,57 @@ var Recv = function () {
     return isIOS || (isAppleDevice && isTouchScreen);
   }
 
+  function _getModeAspectRatio(mode) {
+    // (image_size_x + 16) / (image_size_y + 16)
+    switch (mode) {
+      case 66: return 1.1516; // Bu
+      case 67: return 1.413;  // Bm
+      default: return 1.0;    // B, 4C, auto
```
sz3, line 135: The encoder actually surfaces this information via `cimbare_get_aspect_ratio()` -- we may switch to using that. For now, hardcoding the constants is fine.

web/recv.js
```diff
@@ -119,6 +127,57 @@ var Recv = function () {
...
+    }
+  }
+
+  function _updateCrosshairPositions() {
+    if (!_video || !_video.videoWidth || !_video.videoHeight)
+      return;
+
+    var modeAspect = _getModeAspectRatio(_mode);
+
+    var windowW = window.innerWidth;
```
sz3, line 145: Some convoluted nonsense. It works, probably

web/zstd.js
```diff
@@ -84,8 +75,15 @@ var Zstd = function () {
...
-      _downloadHelper(name, bloburl);
+      var aaa = document.createElement('a');
+      aaa.href = bloburl;
+      aaa.download = name;
+      document.body.appendChild(aaa);
+      aaa.style = 'display: none';
+      aaa.click();
+
       setTimeout(function () {
+        aaa.remove();
```
sz3, line 86: Firefox on ios (+ maybe android?) made a change somewhat recently that clobbers the filename on download. AFAIK it's not fixable using this blob approach. It might be time to revisit the steaming API (see commented out code path in this very file)

PR #162, opened 2026-04-25
Inlining/flatten hints for mobile (android) performance

sz3:
The past approach was to use a blunt hammer and have the android app pass a top level `-mllvm -inline-threshold=$BIGNUMBER` flag to clang, then pray the compiler would do the right thing. It worked surprisingly well. Unfortunately not so much for recent clang releases (android NDK 27+), so I had to track down where flatten/inline actually gives us some benefit.

This seems to work ok. CFC will still use the inline-threshold flag, but it can do it more selectively.

note: the flatten/inline macros are *only* defined for clang at present (GCC has slightly different rules), and the one in CimbReader requires LTO to be enabled.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,6 +7,8 @@ if (MSVC)
     add_compile_options(/FI "iso646.h")
 endif()
 
+set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
+
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
 	set(CMAKE_INSTALL_PREFIX "${libcimbar_SOURCE_DIR}/dist" CACHE PATH "..." FORCE)
 endif()
--- a/src/lib/cimb_translator/CimbDecoder.cpp
+++ b/src/lib/cimb_translator/CimbDecoder.cpp
@@ -6,6 +6,7 @@
 #include "Config.h"
 #include "image_hash/hamming_distance.h"
 #include "serialize/format.h"
+#include "util/compiler_constants.h"
 
 #include <algorithm>
 #include <iostream>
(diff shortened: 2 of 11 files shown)


src/lib/cimb_translator/test/CimbReaderTest.cpp
```diff
@@ -25,6 +25,11 @@ namespace {
 	public:
 		using CimbDecoder::CimbDecoder;
 		using CimbDecoder::internal_ccm;
+
+		~TestableCimbDecoder()
+		{
+			internal_ccm() = color_correction();
```
sz3, line 31: Unrelated: fix testing bug (cleanup, so we don't sabotage other tests)

src/lib/cimb_translator/CimbDecoder.cpp
```diff
@@ -138,7 +139,7 @@ unsigned CimbDecoder::decode_symbol(const cv::Mat& cell, unsigned& drift_offset,
 	return get_best_symbol(results, drift_offset, best_distance, cooldown);
 }
 
-unsigned CimbDecoder::decode_symbol(const bitmatrix& cell, unsigned& drift_offset, unsigned& best_distance, unsigned cooldown) const
+CIMBAR_FLATTEN unsigned CimbDecoder::decode_symbol(const bitmatrix& cell, unsigned& drift_offset, unsigned& best_distance, unsigned cooldown) const
```
sz3, line 142: The hot paths we really care about are decode_symbol() and decode_color(). It's tricky because these are both CPU cache-limited, so it's a moving target depending on what hardware we're trying to optimize for. (probably none of this is news to you if you're familiar with low level optimization)

My claim/thesis here is that it's better to target old hardware, because modern CPUs have more cache and can power through. OTOH, a phone from 2017 needs all the help it can get...

CMakeLists.txt
```diff
@@ -7,6 +7,8 @@ if (MSVC)
     add_compile_options(/FI "iso646.h")
 endif()
 
+set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
```
sz3, line 10: LTO!

sz3: You only need this if you have the inline/flatten macros set (due to CimbReader). So if it causes trouble we can add a clang gate here as well.

PR #164, opened 2026-05-09
web encoder: use requestAnimationFrame(), trim size (slightly) for asm.js (single file) build

sz3:
Small stuff.


src/lib/cimbar_js/CMakeLists.txt
```diff
@@ -44,7 +44,7 @@ target_link_libraries(cimbar_js
 
 # asm.js vs wasm
 if(USE_WASM EQUAL "2")
-	set (LINK_WASM_OPT "-s SINGLE_FILE=1")
+	set (LINK_WASM_OPT "-s SINGLE_FILE=1 --closure 1")
```
sz3, line 47: Debated using emmalloc as well, but I don't know the performance impact.

Issue #149, opened 2025-11-21
Suggest providing a list of recommended hardware


AsianOtter:
This project is fantastic; I've successfully transferred files from my PC to my Android phone.

The only downside is that transferring files larger than 20MB causes the phone to overheat and the app to crash; however, that's okay, I only need to transfer small files.

I'd like to use your project to transfer files between two physically isolated networks, but I'm unsure how to choose the hardware to ensure long-term stability. I imagine the author is the most knowledgeable about the project. Could you recommend a hardware list, including the camera, CPUs, GPUs, and RAM size at both ends? Are there any other considerations?
（If you have the necessary hardware for the design, I would be happy to purchase it.）

sz3:
> The only downside is that transferring files larger than 20MB causes the phone to overheat and the app to crash

That's a funny one I haven't seen before. Though maybe not shocking (the decoder will use all the processing it can)

> I'd like to use your project to transfer files between two physically isolated networks, but I'm unsure how to choose the hardware to ensure long-term stability. I imagine the author is the most knowledgeable about the project. Could you recommend a hardware list, including the camera, CPUs, GPUs, and RAM size at both ends? Are there any other considerations?

Unfortunately, I don't have any specific recommendations -- I haven't looked too hard into setting this up. PC-compatible cameras in particularly are a blindspot (I have a couple but wouldn't recommend them). I've thought about using an old android phone as a camera, but haven't spent a lot of time looking into that.

But I can give some general-purpose engineering thoughts on the problem:
* CPU: the receiver/decoder is the bottleneck. My original test device (which I still use from time to time) is an android phone with a Qualcomm Snapdragon 625 (~2GHz, ARM Cortex-A53). Multiple cores is probably a must, so e.g. a raspberry pi zero might be too low end (but e.g. a Raspberry Pi Zero 2 W is probably fine, though I haven't tested it)
* GPU: helps a bit, but I'd say a better than 10 year old mobile SOC is sufficient. Having a GPU-like interface can be nice because openCV will use e.g. openCL to speed up decoder ops. On the sender/encoder, we draw to the screen with openGL.
    * note: for something like this, you could also use the `cimbar_send` binary instead of the web UI
* RAM: sender and receiver both need enough contiguous memory to hold the compressed data. I don't have a great intuition for where this breaks down, but it's worth noting that this is probably only an issue if the browser is involved. In other words, if we used cimbar_send instead of the js encoder, RAM probably isn't a concern
[...]

PR #165, opened 2026-05-09
docs(README): add Community Tools section linking cimbar-bigfile

### LINKED ISSUE #163: A community tool implementing the "split larger files" idea from DETAILS.md
Hi @sz3,

I came across this note in DETAILS.md:

> "the encoder and decoder could agree on a scheme to split up, and then reassemble, larger files. However, Cimbar does not (yet?) implement this."

I built a pure-frontend wrapper that does exactly this, without requiring any changes to libcimbar or cfc:

**How it works** (purely user-space, no protocol/wasm/cfc changes):
- File is sliced into N chunks client-side (default 10MB)
- Each chunk uses an independent `encode_id`, so cfc's existing `fountain_decoder_sink` handles them as N parallel streams via the standard `unordered_map<encode_id, stream>` bucket
- Per-chunk filenames go through the standard ZSTD header (`_cimbare_init_encode`), so cfc saves chunks as `<basename>.partNN.bin` natively
- A `manifest.json` (sent as the first stream) carries SHA256 hashes
- A zero-dependency `reassemble.html` does concatenation + verification

**Repo**: https://github.com/xPeiPeix/cimbar-bigfile
**Tested**: 100MB application-layer round-trip verified (chunking + manifest + reassembly + full-file SHA256; automated test in `scripts/test-round-trip-100mb.js`). Physical optical link manually verified at ~28MB end-to-end.
**Licensing**: MIT for original code; vendored libcimbar v0.6.4 wasm under MPL-2.0 with source link back to your release; embedded wirehair under BSD-3-Clause. All three licenses documented in README.

If you'd be open to it, I'd be happy to send a small README PR adding a "Community tools" link — but completely understand if you'd rather keep the README minimal. Either way, sharing this in case it's useful for users who hit the 33MB cap.

Thank you for the great work on libcimbar!

xPeiPeix:
Per discussion in #163.

Closes #163.

Adds a new "Community Tools" section to `README.md` with a single one-line entry linking to [cimbar-bigfile](https://github.com/xPeiPeix/cimbar-bigfile) — a pure-frontend wrapper that splits files into multiple parallel `encode_id` streams (10–15 MB chunks per stream, well below the single-stream wirehair capacity ceiling) and ships a `manifest.json` for SHA256-verified reassembly. No changes to libcimbar or cfc required.

**Placement**: between "Room for improvement/next steps" and "Inspiration", since the tool is a natural extension of the TODO note about splitting larger files.

Happy to adjust placement, wording, or formatting per your preference.

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ You can also encode a file using [cimbar.org](https://cimbar.org), or the latest
 
 ## Community Tools
 
-* [cimbar-bigfile](https://github.com/xPeiPeix/cimbar-bigfile) — A pure-frontend wrapper that splits larger files into parallel `encode_id` streams with a `manifest.json` for reassembly, lifting the 33MB single-file cap without any protocol changes.
+* [cimbar-bigfile](https://github.com/xPeiPeix/cimbar-bigfile) — A pure-frontend wrapper that splits files into multiple parallel `encode_id` streams (10–15 MB chunks per stream, well below the single-stream wirehair capacity ceiling) and ships a `manifest.json` for SHA256-verified reassembly. No changes to libcimbar or cfc required.
 
 ## Inspiration
 
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ You can also encode a file using [cimbar.org](https://cimbar.org), or the latest
 
 ## Community Tools
 
-* [cimbar-bigfile](https://github.com/xPeiPeix/cimbar-bigfile) — A pure-frontend wrapper that splits files into multiple parallel `encode_id` streams (10–15 MB chunks per stream, well below the single-stream wirehair capacity ceiling) and ships a `manifest.json` for SHA256-verified reassembly. No changes to libcimbar or cfc required.
+* [cimbar-bigfile](https://github.com/xPeiPeix/cimbar-bigfile) — A pure-frontend wrapper that splits files into multiple parallel `encode_id` streams (10–15 MB chunks per stream), and ships a `manifest.json` for SHA256-verified reassembly. Supports 100+ MB files. No changes to libcimbar or cfc required.
 
 ## Inspiration
 
--- a/README.md
+++ b/README.md
@@ -109,6 +109,10 @@ You can also encode a file using [cimbar.org](https://cimbar.org), or the latest
 
 [TODO](TODO.md)
 
+## Community Tools
+
+* [cimbar-bigfile](https://github.com/xPeiPeix/cimbar-bigfile) — A pure-frontend wrapper that splits larger files into parallel `encode_id` streams with a `manifest.json` for reassembly, lifting the 33MB single-file cap without any protocol changes.
+
 ## Inspiration
 
 * https://github.com/JohannesBuchner/imagehash/


sz3:
What's the biggest file you've been able to send this way?

xPeiPeix:
Hi @sz3, thanks for asking!

The largest file I've personally pushed end-to-end is ~100 MB. With the default params (Mode B / 15 fps / 10 MB chunks / 1.5x redundancy, scanning a 1080p screen with a Pixel 5), the run looks roughly like:

- 10 data chunks + 1 manifest → 11 save dialogs on the CFC side
- ~16–20 minutes wall-clock end-to-end
- ~106 KB/s effective throughput

(Full numbers in [README.en.md → Performance reference](https://github.com/xPeiPeix/cimbar-bigfile/blob/main/README.en.md#performance-reference).)

The theoretical ceiling is bounded by libcimbar's wirehair fountain decoder — encode_id slots get recycled roughly every ~128 transfers (per #149), so a single session's `chunk_count` should stay under ~120, which works out to about **~1.2 GB at 10 MB/chunk** as a practical upper bound. Past that, CFC starts dropping frames whose `encode_id` slot has been reused.

---

And one small UX observation from the multi-chunk side that you may already be aware of:

In CFC's progress UI, switching the on-screen stream doesn't seem to refresh the displayed progress bar. e.g. if I sit on part04 for a while, jump to part05, then come back to part04, the on-screen bar keeps advancing on the part05 progress (the underlying `fountain_decoder_sink` is bucketing both correctly — `cimbard_get_report` exposes per-stream state — it's just the rendering layer that doesn't switch which stream it tracks).

For multi-stream workflows it can make users unsure which stream is actually being received. We worked around it on the sender side (a "mark this chunk saved" button + a "lock current chunk" toggle so the on-screen stream stays put), which keeps the UX usable. Just sharing the observation in case it's useful context.

Anonymous3-a:
I thought the max chunk size was 33MB, which would come out to 3.96GB.

sz3:
> The largest file I've personally pushed end-to-end is ~100 MB

Ok, that's the number I was looking for. I think the logic is sound, I'm just wary about claiming infinite transfer sizes. It's all fun and games until we find out there's some mysterious problem at 500 MB... also, I pity whoever tests that. :slightly_smiling_face: 

A bit of trivia that may be useful for the file size limit, and why lower sizes (e.g. 10MB) tend to be a better choice than the full 33 MB for an individual chunk. wirehair has (IIRC) a uint16_t block_id. The wirehair block size in mode B is 625 bytes (7500/12). (edit: I wrote 600 here. math is hard) Quick math puts that at about ~~39~~ 41 MB of valid blocks, or in other words max 1.2x redundancy for a 33 MB file -- this is a hard cap. The counter will roll over when we hit it, so nothing should break (I hope -- I tested this at one point and it seemed fine), but basically we can ask for as much redundancy as we want but reality limits us.

Trivia: for this reason, at one point early on I had the file size limit at 16 MB, not 33. :upside_down_face: 

Anyway, the combination of uint16_t and block size also means `mode Bu` can't actually transfer a 30 MB file. Its block size is smaller, so its cap is also lower. ...Which I should probably document. In general I think 10-15 MB is a good spot as a chunk size. (there's no penalty for redundant blocks, AFAIK. So feel free to use it all, e.g. 3x or 4x for 10 MB chunks)

One more bit of commentary I'm sure I've made before. Provided you're finished sending a chunk, you can re-use the encode_id if you slightly vary the chunk size... (e.g. 10.01 MB chunks after the first go around) So it takes a lot of baby sitting, but this approach probably can be extended to some obscenely large file sizes. (you can also restart the decoder to clear its cache of "done" files, which will have the same effect without changing the chunk size)

> And one small UX observation from the multi-chunk side that you may already be aware of:
> 
> In CFC's progress UI, switching the on-screen stream doesn't seem to refresh the displayed progress bar. e.g. if I sit on part04 for a while, jump to part05, then come back to part04, the on-screen bar keeps advancing on the part05 progress (the underlying `fountain_decoder_sink` is bucketing both correctly — `cimbard_get_report` exposes per-stream state — it's just the rendering layer that doesn't switch which stream it tracks).
> 
> For multi-stream workflows it can make users unsure which stream is actually being received. We worked around it on the sender side (a "mark this chunk saved" button + a "lock current chunk" toggle so the on-screen stream stays put), which keeps the UX usable. Just sharing the observation in case it's useful context.

That seems like a bug. I'll take a look and see what I can find out.

xPeiPeix:
Hi @sz3, thanks for the detailed trivia — I've folded everything you mentioned into cimbar-bigfile main ([commit 4253901](https://github.com/xPeiPeix/cimbar-bigfile/commit/4253901)):

- **`docs/manifest-spec.md`**: added the wirehair `uint16_t` × 600-byte block-size derivation (~39 MB Mode B hard cap, 33 MB → 1.2x max redundancy as a hard ceiling) + a chunk_size selection table + a note that Mode Bu can't transfer ≥30 MB files
- **`README` (CN + EN)**: corrected the tagline to reference the ~39 MB single-stream cap rather than a vague "10 MB limit"; added a "Choosing a redundancy multiplier" section that quotes your "no penalty for 3x or 4x" line and explains the fountain-code intuition (surplus frames don't waste receive time)
- **`send.html`**: bumped the default redundancy from 1.5x to 2.0x and raised the input max from 3.0x to 5.0x; shifted the three presets up accordingly (balanced 1.5→2.0, aggressive 1.2→1.5, conservative 2.0→3.0)
- Added a new "Advanced: files larger than ~1.2 GB" section documenting the two encode_id re-use tricks you mentioned (vary chunk size between sessions / restart CFC to clear the "done" cache), explicitly marked as manual babysitting since neither is automated yet

Thanks again for the trivia — really helped sharpen both the docs and the defaults. Looking forward to whatever you find on the CFC progress-bar bug.

README.md
```diff
@@ -109,6 +109,10 @@ You can also encode a file using [cimbar.org](https://cimbar.org), or the latest
 
 [TODO](TODO.md)
 
+## Community Tools
+
+* [cimbar-bigfile](https://github.com/xPeiPeix/cimbar-bigfile) — A pure-frontend wrapper that splits files into multiple parallel `encode_id` streams (10–15 MB chunks per stream, well below the single-stream wirehair capacity ceiling) and ships a `manifest.json` for SHA256-verified reassembly. No changes to libcimbar or cfc required.
```
sz3: Maybe:

> parallel `encode_id` streams (10–15 MB chunks per stream), and ships a `manifest.json` for SHA256-verified reassembly. Supports 100+ MB files. No changes to libcimbar or cfc required.

xPeiPeix: Applied — thanks for the cleaner phrasing! Dropped the wirehair jargon and added the "100+ MB" capability anchor in [c0a3566](https://github.com/xPeiPeix/libcimbar/commit/c0a3566).

sz3:
Looks good.

> 600-byte block-size derivation (~39 MB Mode B hard cap, 33 MB → 1.2x max redundancy as a hard ceiling) 

Re-reading this, I did the (very basic) math wrong. :upside_down_face: 

Anyway, the actual numbers are 625 bytes (technically 619 bytes because we have to subtract out the fountain header... but anyway) -> ~40.5 MB.

xPeiPeix:
Ha, thanks for the correction — applied in [9c67371](https://github.com/xPeiPeix/cimbar-bigfile/commit/9c67371). Manifest-spec and README tagline now use ~40.5 MB / 625 bytes per block (619 after subtracting the fountain header).

PR #168, opened 2026-06-13
Fix compiler complaints, clean up GLSL-adjacent code, user newer emsdk, minor API change

sz3:
Miscellaneous tidying up.

* users of the `cimbare_*` apis (e.g. the web encoder) now need to call `cimbare_init_window()` + `cimbare_configure()`, rather than just the latter. Well, *if* you need a window. :slightly_smiling_face: The change should allow usage of the API without a window/GL.
* emsdk 5.0 + opencv 4.12 for the web build.
* tiny 3rd party dep changes: libcorrect (code gate) fix, wirehair (linker tweak -- not wirehair's fault)
* primordial (ok, we can call it "WIP") automated test for the sender. Very basic, but it feels better to have one.
* lots of GLSL refactoring. As I've learned more about shaders, I've learned more about how I'm using shaders wrong. This is more correct, although (having gone through the exercise now), reallocations and not caching GL uniform locations properly has basically nothing to do with the performance issues I've seen inside the PWA or android app. (tl;dr: we need to use offscreencanvas to make those work better. so stay tuned for that follow up PR...)

### DIFF
--- a/src/lib/cimbar_js/cimbar_js.cpp
+++ b/src/lib/cimbar_js/cimbar_js.cpp
@@ -30,7 +30,12 @@ extern "C" {
 
 int cimbare_init_window(int width, int height)
 {
-	// must be divisible by 4???
+	if (width <= 0)
+		width = cimbar::Config::image_size_x() + 16;
+	if (height <= 0)
+		height = cimbar::Config::image_size_y() + 16;
+
+	// GL says must be divisible by 4???
 	if (width % 4 != 0)
 		width += (4 - width % 4);
 	if (height % 4 != 0)
@@ -194,16 +199,8 @@ int cimbare_encode(const unsigned char* buffer, unsigned size)
 
 int cimbare_configure(int mode_val, int compression)
 {
-	cimbar::Config::update(mode_val);
 	if (compression < 0 or compression > 22)
 		compression = cimbar::Config::compression_level();
-
-	// make sure we've initialized
-	int window_size_x = cimbar::Config::image_size_x() + 16;
-	int window_size_y = cimbar::Config::image_size_y() + 16;
-	int initRes = cimbare_init_window(window_size_x, window_size_y);
-	if (initRes < 0)
-		return initRes;
 
 	// check if we need to refresh the stream
 	bool refresh = (mode_val != _modeVal or compression != _compressionLevel);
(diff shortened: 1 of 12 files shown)


src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -214,19 +211,29 @@ int cimbare_configure(int mode_val, int compression)
...
+		if (_window)
+		{
+			int initRes = cimbare_init_window(0, 0);
+			if (initRes < 0)
+				return initRes;
+		}
+
 		// try to refresh the stream
-		if (_window and _fes)
+		if (_fes)
```
sz3, line 223: We want the C APIs to do the right thing if the window is there, but not be dependent on it.

src/lib/gui/gl_2d_display.h
```diff
@@ -11,6 +12,77 @@
 
 namespace cimbar {
 
+class gl_2d_display_program
```
sz3, line 15: Refactor of how the shader code is organized.

I spent some time chasing performance optimizations of the encoder/send code, including a version of the encoder that runs entirely on the GPU. It turns out very little of that is necessary (the slowness is entirely attributable to main thread stall/contention issues on mobile), but the process did convince me that this existing tiny little shader needs some refactoring.

src/lib/gui/gl_2d_display.h
```diff
@@ -45,13 +117,26 @@ class gl_2d_display
...
 	{
-		glGenBuffers(3, _vbo.data());
-		glGenVertexArrays(1, &_vao);
+		init();
+	}
+
+	~gl_2d_display()
+	{
+		if (_texid)
+			glDeleteTextures(1, &_texid);
```
sz3, line 130: e.g. maybe we should clean up our resources properly. 😖

src/lib/gui/gl_program.h
```diff
@@ -10,9 +10,10 @@ namespace cimbar {
 class gl_program
 {
 public:
-	gl_program(GLuint vertexShader, GLuint fragmentShader, const std::string& vertextVarName)
+	template<typename... StringList>
+	gl_program(GLuint vertexShader, GLuint fragmentShader, StringList... vertextVarName)
```
sz3, line 14: Trying out C++17 param expansion. We could also pass a lambda that specifies glBindAttribLocation() calls from the caller. In some ways that (the lambda) makes for more readable code, so I might end up settling on that way instead. Nevertheless: this works for now

src/third_party_lib/libcorrect/src/reed-solomon/polynomial.c
```diff
@@ -212,6 +212,9 @@ polynomial_t polynomial_init_from_roots(field_t field, unsigned int nroots, fiel
 
 polynomial_t polynomial_create_from_roots(field_t field, unsigned int nroots, field_element_t *roots) {
     polynomial_t poly = polynomial_create(nroots);
+    if (nroots == 0) // SZ3: we do a bit of defensive programming, as a treat
```
sz3, line 215: Fix compiler warning.

src/third_party_lib/wirehair/CMakeLists.txt
```diff
@@ -1,6 +1,7 @@
 cmake_minimum_required(VERSION 3.5)
 project(wirehair)
 
+set(CMAKE_INTERPROCEDURAL_OPTIMIZATION FALSE)
```
sz3, line 4: turn off LTO for wirehair.

PR #171, opened 2026-06-28
Add ?ww=1 param to web encoder to opt into encoding on background thread

sz3:
On most platforms (that I've tested anyway), the encoder (+GL calls) runs fast enough that staying on the main thread is fine. (Trivia: on mobile platforms under normal circumstances, the GL calls get offloaded from the main thread automatically)

But in two specific cases we see real slowdown (can't hit 15 fps, have to downgrade to 10 or lower):
1. android PWA operating in offline mode. I don't know why there's a difference, and this probably varies by android version
2. android app webview, a.k.a. what CFC currently uses to embed the encoder.

So: if a param is supplied (either `?ww=1` in the query string, or `#ww=1` as the fragment), we spin up a webworker and hand it the canvas (via OffscreenCanvas) to run in the background.

Additionally:
* fixed a bug in our pwa caching. Oops.
* simple automated browser test for encoder JS logic. Long overdue.

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -375,20 +375,19 @@
   <script src="main.js"></script>
 
   <script type="text/javascript">
-    var canvas = document.getElementById('canvas');
-    Main.init_ww(canvas);
-
     if ('serviceWorker' in navigator) {
       navigator.serviceWorker.register('./sw.js');
     }
 
-    var Module = {};
-    Module.canvas = canvas;
-    Module.onRuntimeInitialized = () => {
-      Main.init(canvas);
-    };
+    var canvas = document.getElementById('canvas');
+    if (!Main.init_ww(canvas)) {
+      var Module = {};
+      Module.canvas = canvas;
+      Module.onRuntimeInitialized = () => {
+        Main.init(canvas);
+      };
+    }
   </script>
-
   <script src="cimbar_js.js"></script>
 
 </body>
--- a/web/index.html
+++ b/web/index.html
@@ -371,11 +371,16 @@
     </div>
   </div>
 
+  <script src="send.js"></script>
+  <script src="main.js"></script>
+
   <script type="text/javascript">
+    var canvas = document.getElementById('canvas');
+    Main.init_ww(canvas);
+
     if ('serviceWorker' in navigator) {
       navigator.serviceWorker.register('./sw.js');
     }
-    var canvas = document.getElementById('canvas');
 
     var Module = {};
     Module.canvas = canvas;
@@ -385,8 +390,6 @@
   </script>
 
   <script src="cimbar_js.js"></script>
-  <script src="send.js"></script>
-  <script src="main.js"></script>
 
 </body>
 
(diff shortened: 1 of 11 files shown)


web/test/run-in-browser.js
```diff
@@ -6,12 +6,18 @@ const puppeteer = require('puppeteer');
 const path = require('path');
 const QUnit = require('qunit');
 
-(async () => {
-  const browser = await puppeteer.launch({ headless: true, args: ['--no-sandbox', '--disable-setuid-sandbox'] });
+const testPages = ['test_recv.html', 'test_send.html'];
```
sz3, line 9: Now there are 2 tests... no automated offscreencanvas test yet, alas

web/index.html
```diff
@@ -371,22 +371,24 @@ <h4>Framerate</h4>
...
-    var canvas = document.getElementById('canvas');
 
-    var Module = {};
-    Module.canvas = canvas;
-    Module.onRuntimeInitialized = () => {
-      Main.init(canvas);
-      Main.nextFrame();
-    };
+    var canvas = document.getElementById('canvas');
+    if (!Main.init_ww(canvas)) {
```
sz3, line 383: init_ww() will early return if neither of the fragment/param is set.

sz3: Notably, the default being this way means we will always load the wasm on the main thread, *even if we don't need it*. This is a tradeoff, but given that most platforms seem to handle it fine, my thought is that "just use the main thread" is the better (simpler) default.

web/main.js
```diff
@@ -1,21 +1,61 @@
+var Report = function () {
```
sz3, line 1: Interface to let us use the same code on main thread/worker thread. There are surely cleaner ways to do this, but it seems ok.

web/send-worker.js
```diff
@@ -0,0 +1,112 @@
+'use strict';
+
+// polyfill
+if (typeof window === 'undefined') {
+  self.window = self;
+}
+
+// stubbing extra things out for emscripten's GLFW
```
sz3, line 8: All these polyfills are pretty gross/hacky and by far my least favorite part of this change.

web/send.js
```diff
@@ -0,0 +1,174 @@
+
+var Send = function () {
```
sz3, line 2: Same code, called either from the main thread or via the worker.

web/sw.js
```diff
@@ -25,19 +25,25 @@ self.addEventListener('install', function (e) {
...
+    caches.match(e.request, { ignoreSearch: true }).then(function (response) {
       return response || fetch(e.request);
     })
   );
 });
 
 // clean old caches
 self.addEventListener('activate', function (e) {
-  e.waitUntil(function () {
+  e.waitUntil(
```
sz3, line 36: So, this is unfortunate (and embarrassing, perhaps) -- we weren't cleaning up old PWA caches properly. I knew something was off with the cache cleanup, but I didn't know what it was... hopefully this way actually works.

PR #172, opened 2026-06-29
relax the C api's cimbare_encode() bufize check to allow multiples of chunk size (16kb)

sz3:
There's no reason not to allow this, and it gives us more flexibility downstream (ex: maybe the caller wants to use 128k buffers instead of 16k).

### DIFF
--- a/src/lib/cimbar_js/cimbar_js.cpp
+++ b/src/lib/cimbar_js/cimbar_js.cpp
@@ -178,7 +178,7 @@ int cimbare_encode(const unsigned char* buffer, unsigned size)
 		if (!_comp->write(reinterpret_cast<const char*>(buffer), size))
 			return -2;
 	}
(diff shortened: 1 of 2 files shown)


src/lib/cimbar_js/cimbar_js.cpp
```diff
@@ -178,7 +178,7 @@ int cimbare_encode(const unsigned char* buffer, unsigned size)
 		if (!_comp->write(reinterpret_cast<const char*>(buffer), size))
 			return -2;
 	}
-	if (size == cimbare_encode_bufsize())
+	if (size%cimbare_encode_bufsize() == 0 and size != 0)
```
sz3, line 181: Nothing too complex here -- the interesting part is in the (unchanged) zstd_compressor::write() function.

This (relaxed) constraint is a bit of a leaky abstraction, but it's convenient so we'll deal with it.

PR #174, opened 2026-07-12
cimbar send js fixes

sz3:
Follow up to the (still unreleased) #171 and #172.

* fix the `force_local` redirect/fallback for the new offscreencanvas+ww impl (for android webview and PWA). Tested on an assortment of devices...
* increase the buff size to 256kb (16kb zstd chunk size * 16) for the web encoder's `importFile()` (when we loop over input file bytes). The buff size was shrunk (too) conservatively in #145 -- 256kb of contiguous memory shouldn't be too much to ask for I think. In some circumstances, e.g. when used in the webview on my older android phone, the smaller buffer creates a *very* noticeable performance hit on file load. This speeds things back up.

### DIFF
--- a/web/main.js
+++ b/web/main.js
@@ -101,9 +101,8 @@ var Main = function () {
       if (force_local) {
         _ww = undefined;
         // force refresh in main thread mode
-        window.location.hash = "";
-        window.location.search = "";
-        window.location.reload();
+        const cleanURL = window.location.origin + window.location.pathname;
+        window.location.replace(cleanURL);
         return;
       }
 
--- a/web/send.js
+++ b/web/send.js
@@ -33,7 +33,7 @@ var Send = function () {
   }
 
   function importFile(file) {
-    let chunkSize = Module._cimbare_encode_bufsize();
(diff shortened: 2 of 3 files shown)


web/main.js
```diff
@@ -101,9 +101,8 @@ var Main = function () {
       if (force_local) {
         _ww = undefined;
         // force refresh in main thread mode
-        window.location.hash = "";
-        window.location.search = "";
-        window.location.reload();
+        const cleanURL = window.location.origin + window.location.pathname;
+        window.location.replace(cleanURL);
```
sz3, line 105: Setting them individually creates a redirect loop 🙃

web/send.js
```diff
@@ -33,7 +33,7 @@ var Send = function () {
   }
 
   function importFile(file) {
-    let chunkSize = Module._cimbare_encode_bufsize();
+    let chunkSize = Module._cimbare_encode_bufsize() * 16;
```
sz3, line 36: #172 relaxes the check -- this uses it.

Could make this `16` configurable, but I'm hoping it won't cause any trouble as is.

PR #175, opened 2026-07-13
decompression loop fix + cleanup for decode hot path + overdue unit test for bitmatrix

sz3:
* a specially constructed zstd frame could cause an infinite loop (-> crash the app). Oops
* attempts at optimizing bitmatrix (under the theory we were cache-limited) proved unsuccessful. But there are some small optimizations and bits of cleanup that fell out.
* added a unit test for bitmatrix... we were previously only testing it one level up as part of the decode.

### DIFF
--- a/src/lib/bit_file/bitbuffer.h
+++ b/src/lib/bit_file/bitbuffer.h
@@ -1,6 +1,7 @@
 /* This code is subject to the terms of the Mozilla Public License, v.2.0. http://mozilla.org/MPL/2.0/. */
 #pragma once
 
+#include "util/compiler_constants.h"
 #include <vector>
 
 // write bits -> buffer
@@ -16,20 +17,25 @@ public:
 		{}
 
 		template <typename UINT>
-		bool write_byte(UINT byte)
+		inline bool write_byte(UINT byte)
 		{
-			return write(byte, 8);
+			size_t pos = _pos;
+			_pos += 8;
+			if (pos % 8 != 0)
+				return _bb.write(byte, pos, 8);
+			// else
+			return _bb.write_aligned_byte(byte, pos);
 		}
 
 		template <typename UINT>
-		bool write(UINT byte, int length)
+		inline bool write(UINT byte, int length)
 		{
 			size_t pos = _pos;
 			_pos += length;
 			return _bb.write(byte, pos, length);
 		}
 
-		writer& operator<<(uint8_t byte)
+		inline writer& operator<<(uint8_t byte)
 		{
 			write_byte(byte);
 			return *this;
@@ -42,12 +48,12 @@ public:
 
 public:
 	bitbuffer(unsigned size_hint=9300)
-	    : _sizeHint(size_hint)
+		: _sizeHint(size_hint)
 	{
 		clear();
 	}
 
-	void resize(unsigned length)
+	inline void resize(unsigned length)
 	{
 		size_t requestedSize = length/8;
 		size_t size = _buffer.size();
@@ -59,7 +65,7 @@ public:
 		_buffer.resize(size, 0);
 	}
 
-	bool write(unsigned data, unsigned index, int length)
+	inline bool write(unsigned data, unsigned index, int length)
 	{
 		resize(index+length);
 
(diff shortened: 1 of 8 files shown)


src/lib/bit_file/test/bitmatrixTest.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+TEST_CASE( "bitmatrixTest/test_mat_to_bitbuffer", "[unit]" )
+{
+	cv::Mat symbols = TestCimbar::loadSample("b/tr_0.png");
+	assertEquals( 1024, symbols.cols );
+	assertEquals( 1024, symbols.rows );
+
+	// grayscale and threshold, to imitate the decoder process
+	cv::cvtColor(symbols, symbols, cv::COLOR_RGB2GRAY);
+	cv::adaptiveThreshold(symbols, symbols, 255, cv::ADAPTIVE_THRESH_MEAN_C, cv::THRESH_BINARY, 7, 0);
```
sz3, line 17: Imitating what we do in CimbReader.

src/lib/bit_file/bitbuffer.h
```diff
@@ -16,20 +17,25 @@ class bitbuffer
...
-		bool write_byte(UINT byte)
+		inline bool write_byte(UINT byte)
 		{
-			return write(byte, 8);
+			size_t pos = _pos;
+			_pos += 8;
+			if (pos % 8 != 0)
+				return _bb.write(byte, pos, 8);
+			// else
+			return _bb.write_aligned_byte(byte, pos);
```
sz3, line 27: While the general case is nice, `mat_to_bitbuffer()` (what we care about) is *always* byte aligned.

src/lib/bit_file/bitbuffer.h
```diff
@@ -83,38 +89,49 @@ class bitbuffer
...
-			currentBit += nextRead;
-			if (currentBit >= 8)
-				currentBit = 0;
-			currentByte += 1;
-			nextRead = std::min(length, 8-currentBit);
+			length -= rdsz;
+			currentBit = 0;
+			++currentByte;
+
+			UINT temp = bits;
```
sz3, line 120: Ultimately this doesn't *need* to be a template, since the code that was going to use it with a different type probably won't get merged. But it's not too onerous (we have to upcast from uint8_t regardless)

src/lib/cimb_translator/CimbDecoder.cpp
```diff
@@ -100,9 +100,11 @@ bool CimbDecoder::load_tiles()
 
 unsigned CimbDecoder::get_best_symbol(image_hash::ahash_result<cimbar::Config::cell_size()>& results, unsigned& drift_offset, unsigned& best_distance, unsigned cooldown) const
 {
-	drift_offset = 0;
 	unsigned best_fit = 0;
+	drift_offset = 0;
 	best_distance = 1000;
+	cooldown = (cooldown == 4)? 0xFF : cooldown;  // don't skip the center, obvs
```
sz3, line 106: It just reads better to move this out of the loop.

src/lib/cimb_translator/FloodDecodePositions.cpp
```diff
@@ -103,12 +103,9 @@ int FloodDecodePositions::update(unsigned index, const CellDrift& drift, unsigne
 		{
 			std::array<int,4> horizon = {-1, -1, -1, -1};
 			horizon[0] = _cellFinder.right(rridx);
-			if (horizon[0] >= 0)
-				horizon[1] = _cellFinder.right(horizon[0]);
+			horizon[1] = _cellFinder.right(horizon[0]);
```
sz3, line 106: cell finder does bounds checks (it didn't originally), so we can be lazy here.

src/lib/cimb_translator/LinearDecodePositions.h
```diff
@@ -13,13 +13,15 @@ class LinearDecodePositions
 public:
 	LinearDecodePositions(cimbar::vec_xy spacing, cimbar::vec_xy dimensions, int offset, cimbar::vec_xy marker_size);
 
-	size_t count() const;
+	size_t size() const;
```
sz3, line 16: For profiling purposes, I wanted to ditch the flood decode temporarily-- which meant getting this class working again.

The interface doesn't exactly match (flood decode has an extra param it tracks), but it doesn't need to be drop-in compatible to be (occasionally) useful.

src/lib/compression/zstd_decompressor.h
```diff
@@ -46,7 +47,7 @@ class zstd_decompressor : public STREAM
 			}
 			_inBuff = std::string_view(_inBuff.data() + input.size, _inBuff.size() - input.size);
 		}
-		return false;
+		return true;
```
sz3, line 50: We'll now treat this as "should we continue?"

sz3:
Been testing all of this for a couple weeks now, so it should be solid.

PR #176, opened 2026-07-13
fix package-cimbar-html.js

sz3:
whoopsies

Fallout from #171.


package-cimbar-html.py
```diff
@@ -21,9 +22,11 @@ def read_script(name):
 
 def main():
     contents = read_file('index')
+    send_js = read_script('send_js')
     main_js = read_script('main_js')
     cimbar_js = read_script('cimbar_js')
 
+    contents = contents.replace('<script src="send.js">', send_js)
```
sz3, line 29: Nothing complicated.

PR #177, opened 2026-07-14
Fix PWAs for offline operation

sz3:
Some last-second v0.6.7 (https://github.com/sz3/libcimbar/releases/tag/v0.6.7c) changes for cimbar.org. :upside_down_face:

### DIFF
--- a/web/pwa-recv.json
+++ b/web/pwa-recv.json
@@ -5,7 +5,7 @@
 	"short_name": "cimbar recv",
 	"icons": [{"src":"icon-192x192.png","sizes":"192x192","type":"image/png"},{"src":"icon-512x512.png","sizes":"512x512","type":"image/png","purpose": "any"},{"src":"icon-512x512-maskable.png","sizes":"512x512","type":"image/png","purpose": "maskable"}],
 	"scope": "/",
-	"start_url": "recv.html",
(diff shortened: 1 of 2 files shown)


web/pwa-recv.json
```diff
@@ -5,7 +5,7 @@
 	"short_name": "cimbar recv",
 	"icons": [{"src":"icon-192x192.png","sizes":"192x192","type":"image/png"},{"src":"icon-512x512.png","sizes":"512x512","type":"image/png","purpose": "any"},{"src":"icon-512x512-maskable.png","sizes":"512x512","type":"image/png","purpose": "maskable"}],
 	"scope": "/",
-	"start_url": "recv.html",
+	"start_url": "/recv.html",
```
sz3, line 8: Offline mode isn't working without this. I was sure it was working before... but that seems like life with PWAs

web/sw.js
```diff
@@ -3,6 +3,7 @@ var _cacheName = 'cimbar-js-v%VERSION%';
 var _cacheFiles = [
   '/',
   '/index.html',
+  '/index.html?ww=1',
```
sz3, line 6: oopsie

(technically I think this means it gets cached twice. Probably fine; it's small)

PR #181, opened 2026-08-10
add windows compile compatibility

eliiran1231:
hey there i noticed there are some issues with compiling the sender, reciver, etc for windows
this pr solves these to ensure windows can produce exes without harming the linux compatibility

PR #182, opened 2026-08-14
CI for (beta) windows MSVC support

sz3:
Building on #181. Add windows CI in github actions. Not *currently* planning on official windows builds, but it's nice to have for development purposes, and MSVC brings a different angle to compilation than g++/clang (which can be more annoying than useful.. but sometimes it's useful)

Anyway, the main trick/pain here is vcpkg caching.

*also*, this will momentarily break `master` CI checks, because there are some fixes that windows needs. Followup PR will appear shortly (https://github.com/sz3/libcimbar/actions/runs/31844154756) ... and here it is: #183

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,30 +27,56 @@ jobs:
           env: CXX="clang++" CC="clang"
           extra-packages: "clang"
 
+        - name: "windows msvc"
+          os: windows-latest
+
     steps:
     - uses: actions/checkout@v4
       with:
         submodules: 'recursive'
 
-    - name: Install dependencies
+    - name: Install dependencies (linux)
+      if: runner.os == 'Linux'
       run: sudo apt-get update && sudo apt-get install libopencv-dev libglfw3-dev libgles2-mesa-dev ${{ matrix.config.extra-packages }}
 
     - name: Create Build Environment
-      run: cmake -E make_directory ${{runner.workspace}}/build
+      run: |
+        cmake -E make_directory ${{runner.workspace}}/build
+        cmake -E make_directory ${{github.workspace}}/vcpkg-cache
 
-    - name: Configure CMake
+    - name: setup vcpkg cache (windows)
+      if: runner.os == 'Windows'
+      uses: actions/cache@v4
+      with:
+        path: ${{ github.workspace }}/vcpkg-cache
+        key: vcpkg-cache-${{ runner.os }}-${{ hashFiles('vcpkg.json') }}
+        restore-keys: |
+          vcpkg-cache-${{ runner.os }}-
+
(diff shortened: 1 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -27,30 +27,56 @@ jobs:
...
-    - name: Install dependencies
+    - name: Install dependencies (linux)
+      if: runner.os == 'Linux'
       run: sudo apt-get update && sudo apt-get install libopencv-dev libglfw3-dev libgles2-mesa-dev ${{ matrix.config.extra-packages }}
 
     - name: Create Build Environment
-      run: cmake -E make_directory ${{runner.workspace}}/build
+      run: |
+        cmake -E make_directory ${{runner.workspace}}/build
+        cmake -E make_directory ${{github.workspace}}/vcpkg-cache
```
sz3, line 45: No harm in always making the dir

.github/workflows/ci.yml
```diff
@@ -27,30 +27,56 @@ jobs:
...
+      if: runner.os == 'Linux'
       working-directory: ${{runner.workspace}}/build
       run: ${{ matrix.config.env }} cmake $GITHUB_WORKSPACE
 
+    - name: Configure CMake (windows)
+      if: runner.os == 'Windows'
+      env:
+        VCPKG_BINARY_SOURCES: "clear;files,${{ github.workspace }}/vcpkg-cache,readwrite"
+      working-directory: ${{runner.workspace}}/build
+      run: cmake $env:GITHUB_WORKSPACE -A x64 "-DCMAKE_TOOLCHAIN_FILE=$env:VCPKG_INSTALLATION_ROOT/scripts/buildsystems/vcpkg.cmake"
```
sz3, line 66: Idea here is that this should handle the dependency builds (vcpkg). Still less than ideal, since apparently github caches don't get shared between branches (with the hopeful "trust me bro, it'll work" that maybe master/main branch caches are shared??).

Anyway, please be exited for 70 minute compile times

vcpkg.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "libcimbar",
+  "version-string": "0.6.7c",
+  "builtin-baseline": "9e593bb18ea69cc5095e012465dcd675a822ed0d",
```
sz3, line 4: https://github.com/microsoft/vcpkg/releases/tag/2026.07.29

vcpkg.json
```diff
@@ -0,0 +1,13 @@
+{
+  "name": "libcimbar",
+  "version-string": "0.6.7c",
```
sz3, line 3: Not sure if changing this will invalidate the cache or not. Hopefully not. Stay tuned...

PR #184, opened 2026-08-16
Reproducible wasm build + shader downgrade to GLES 100

sz3:
Three parts:
* reproducible wasm build! This turns out to be *fairly* straightforward now. Mostly timestamp things (use SOURCE_DATE_EPOCH + make sure any tar/etc operations use a fixed timestamp), but also depends on the docker image used for the build.
    * caveat: depends on the emsdk docker image. We're currently tracking the "5.0.0" tag in the CI scripts.
* downgrade shader to use `es 100` -- turns out `es 300` is not necessary, which *maybe* will allow some older devices to run the web encoder?
* minor change to PWA json to allow send/recv to coexist. Sharing the same domain seems intractable (unfortunately), it looks like you have to install the recv app *first* to for both installs to work. ... probably wouldn't have done it this way if I'd known.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,10 +7,6 @@ endif()
 project ( libcimbar )
 enable_testing()
 
-if (MSVC)
-	add_compile_options(/FI "iso646.h" /utf-8)
-endif()
-
 set(CMAKE_INTERPROCEDURAL_OPTIMIZATION TRUE)
 
 if(CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
@@ -19,6 +15,14 @@ endif()
 
 set(CMAKE_CXX_STANDARD 17)
 add_definitions("-DLIBCIMBAR_PROJECT_ROOT=\"${libcimbar_SOURCE_DIR}\"")
+
+# reproducible build, hopefully
+if(NOT MSVC)
+	add_compile_options(
+		-ffile-prefix-map=${CMAKE_SOURCE_DIR}=.
+		-ffile-prefix-map=${CMAKE_BINARY_DIR}=.
+	)
+endif()
 
 if("${CMAKE_BUILD_TYPE}" STREQUAL "")
 	set(CMAKE_BUILD_TYPE "RelWithDebInfo")
@@ -70,7 +74,10 @@ if(NOT DEFINED CPPFILESYSTEM)
 	set(CPPFILESYSTEM "stdc++fs")
 endif()
 
+# special snowflake MSVC section
 if(MSVC)
+	add_compile_options(/FI "iso646.h" /utf-8)
+
 	set(CPPFILESYSTEM "")
 endif()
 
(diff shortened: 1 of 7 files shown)


src/lib/cimbar_js/CMakeLists.txt
```diff
@@ -87,8 +87,6 @@ endif()
 set (LINK_WASM_LIST
 	--bind -Os
 	${LINK_WASM_OPT}
-	-s USE_WEBGL2=1
```
sz3, line 90: This is the main benefit of the older GLSL version.

src/lib/gui/gl_2d_display.h
```diff
@@ -15,26 +15,25 @@ namespace cimbar {
...
+	static constexpr const char* VERTEX_SHADER_SRC = R"(
 	uniform mat2 rot;
 	uniform vec2 tform;
-	in vec4 vert;
-	out vec2 texCoord;
+	attribute vec4 vert;
+	varying vec2 texCoord;
 	void main() {
-	   gl_Position = vec4(vert.x, vert.y, 0.0f, 1.0f);
+	   gl_Position = vec4(vert.x, vert.y, 0.0, 1.0);
```
sz3, line 24: Really, nothing too scary.

web/pwa-recv.json
```diff
@@ -4,7 +4,7 @@
 	"name": "cimbar.org Receiver",
 	"short_name": "cimbar recv",
 	"icons": [{"src":"icon-192x192.png","sizes":"192x192","type":"image/png"},{"src":"icon-512x512.png","sizes":"512x512","type":"image/png","purpose": "any"},{"src":"icon-512x512-maskable.png","sizes":"512x512","type":"image/png","purpose": "maskable"}],
-	"scope": "/",
+	"scope": "/recv.html",
```
sz3, line 7: Specifying the more specific scope here *should* let side-by-side installs work properly.

However, it's awkward because on chrome-based browsers, you have to install `recv` first. Will either have to move it to its own subdomain (re.cimbar.org is currently a redirect), or combine the apps... eventually.

CMakeLists.txt
```diff
@@ -20,6 +16,14 @@ endif()
 set(CMAKE_CXX_STANDARD 17)
 add_definitions("-DLIBCIMBAR_PROJECT_ROOT=\"${libcimbar_SOURCE_DIR}\"")
 
+# reproducible build, hopefully
+if(NOT MSVC)
+	add_compile_options(
+		-ffile-prefix-map=${CMAKE_SOURCE_DIR}=.
+		-ffile-prefix-map=${CMAKE_BINARY_DIR}=.
```
sz3, line 23: Haven't done *too* much work on reproducibility for non-wasm builds. Obviously it's mostly the same stuff, but there are probably some extra gotchas.

package-wasm.sh
```diff
@@ -1,6 +1,8 @@
 #!/bin/bash
-#docker run --mount type=bind,source="$(pwd)",target="/usr/src/app" -it emscripten/emsdk:5.0.0 bash
+#docker run -e SOURCE_DATE_EPOCH --mount type=bind,source="$(pwd)",target="/usr/src/app" -it emscripten/emsdk:5.0.0 bash
```
sz3, line 2: you don't *have* to pass `-e SOURCE_DATE_EPOCH` if you don't care, of course.

sz3: Note that the CI build does *not* define SOURCE_DATE_EPOCH (no changes to the github actions script in this PR). We could lock to git commit timestamp, but it seems simpler to let the CI server define the timestamp (e.g. the release bits use the time when the CI server ran the build).

PR #185, opened 2026-08-16
security hardening: small fountain_decoder fixes for bad inputs

sz3:
These fall into the category of theoretical attacks -- but there were some assumptions baked in about inputs that aren't necessarily true (or won't be if the code is rearranged slightly).

These were inspired by #178 , which was presumably created accidentally and not really mergeable in that state anyway.  But +1 for AI security hardening I suppose

### DIFF
--- a/src/lib/fountain/fountain_decoder_sink.h
+++ b/src/lib/fountain/fountain_decoder_sink.h
@@ -54,7 +54,7 @@ class fountain_decoder_sink
 {
 public:
 	fountain_decoder_sink(unsigned chunk_size, const std::function<std::string(const std::string&, const std::vector<uint8_t>&)>& on_store=nullptr)
-		: _chunkSize(chunk_size)
+		: _chunkSize(chunk_size < 50? 50 : chunk_size)
 		, _onStore(on_store)
 	{
 	}
@@ -132,15 +132,18 @@ public:
 
 	int64_t decode_frame(const char* data, unsigned size)
 	{
-		if (size < FountainMetadata::md_size)
+		if ((size%_chunkSize) > 0)
 			return -10;
+
+		if (size < FountainMetadata::md_size)
+			return -11;
 
 		FountainMetadata md(data, size);
 		if (!md.file_size())
 		{
 			/*std::cout << fmt::format("decode frame {} ... {},{},{},{}",
 									 md.file_size(), (unsigned)data[0], (unsigned)data[1], (unsigned)data[2], (unsigned)data[3]) << std::endl;*/
-			return -11;
+			return -12;
 		}
 
 		// check if already done
@@ -151,7 +154,7 @@ public:
 		auto p = _streams.try_emplace(stream_slot(md), md.file_size(), _chunkSize);
 		fountain_decoder_stream& s = p.first->second;
 		if (s.data_size() != md.file_size())
-			return -12;
+			return -13;
 
 		bool finished = s.write(data, size);
 		if (!finished)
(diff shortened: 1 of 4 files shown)


src/lib/fountain/test/fountain_sinkTest.cpp
```diff
@@ -143,3 +143,62 @@ TEST_CASE( "FountainSinkTest/testSameFrameManyTimes", "[unit]" )
 	assertEquals( "0.333333", turbo::str::join(sink.get_progress()) ); // 33% done
 	assertEquals( "", turbo::str::join(sink.get_done()) );
 }
+
+TEST_CASE( "FountainSinkTest/testRejection", "[unit]" )
```
sz3, line 147: tests for ... mostly stuff that already existed, and should've been tested but wasn't. 🙃

src/lib/fountain/test/fountain_sinkTest.cpp
```diff
@@ -143,3 +143,62 @@ TEST_CASE( "FountainSinkTest/testSameFrameManyTimes", "[unit]" )
...
+
+		// extract and clobber
+		FountainMetadata md(buff.data(), FountainMetadata::md_size);
+		md = FountainMetadata(md.encode_id(), 0, md.block_id());
+		std::memcpy(buff.data(), md.data(), FountainMetadata::md_size);
+
+		assertEquals( -11, sink.decode_frame(buff.data(), buff.size()) );
+	}
+
+	// different size, same encode_id
```
sz3, line 192: but this is new

src/lib/fountain/fountain_decoder_sink.h
```diff
@@ -54,7 +54,7 @@ class fountain_decoder_sink
 {
 public:
 	fountain_decoder_sink(unsigned chunk_size, const std::function<std::string(const std::string&, const std::vector<uint8_t>&)>& on_store=nullptr)
-		: _chunkSize(chunk_size)
+		: _chunkSize(chunk_size < 50? 50 : chunk_size)
```
sz3, line 57: 50 is completely arbitrary. It just doesn't make sense to allow chunk sizes below `7` (fountain header is `6` bytes), which begs the question "what's a reasonable minimum?".

As a reference point, `mode B` uses `625`.

src/lib/fountain/fountain_decoder_sink.h
```diff
@@ -132,7 +132,8 @@ class fountain_decoder_sink
 
 	int64_t decode_frame(const char* data, unsigned size)
 	{
-		if (size < FountainMetadata::md_size)
+		// 2nd clause is implicitly true iff 1st is... but leave it for clarity
+		if ((size%_chunkSize) > 0 or size < FountainMetadata::md_size)
 			return -10;
```
sz3, line 137: .

The fix here is that size needs to be a multiple of chunkSize. In practice, it always will be because `aligned_stream` will make sure of that. But it's good to have a sanity check.

src/lib/fountain/fountain_decoder_stream.h
```diff
@@ -46,16 +47,13 @@ class fountain_decoder_stream
 	{
 		// if we're full
 		_buffIndex = 0;
-		// we ignore the first 4 bytes. It's the sink's job to make sure we're getting the right stuff.
-		// we may, at some point, sanity check if data_size == [1]+[2]+[3]
-		unsigned blockId = (unsigned)(_buffer[4]) << 8 | _buffer[5];
-		return _decoder.decode(blockId, _buffer.data() + _headerSize, block_size());
+		FountainMetadata md(reinterpret_cast<const char*>(_buffer.data()), 6);
+		if (data_size() != md.file_size())
+			return false; // sanity check
```
sz3, line 52: .

Issue #180, opened 2026-07-27
Accessibility: Limit to 3 Frames per Second


poyribn:
Every Framerate selectable on cimbar.org is too fast.

To avoid Risk of Seizures from flashing Lights, the Code shouldn’t flash more than 3 Times per Second.

So the seletable Framerates on cimbar.org should be:

- 1 FpS
- 2 FpS
- 3 FpS

This Change also makes the Framerate Selection more discrete,
so it becomes better to display it as Radio Buttons instead of a Slider.

sz3:
Restricting the framerate will cut the data transfer rate to an essentially useless number.

Worth noting that you don't have to stare at the bar code, only the cell phone camera does. In fact I think it's better not to watch too closely once it's centered, especially for larger transfers. (just from a "watched pot never boils" perspective)

We could probably add a seizure warning about staring at "TV static" flashing lights.

poyribn:
> Restricting the framerate will cut the data transfer rate to an essentially useless number.

The Sender and Receiver should decide if that’s true.

Small and optimized Files would still be super fast to obtain compared to an online Download.
I am already optimizing my Files before sending them via Cimbar, so it shouldn’t be an Issue.

My new Idea is to set the Framerate to 3 by Default, unlocking higher Framerates after accepting a seizure Waring:
- 4
- 5
- 6
- 9
- 10
- 12
- 15
- 20
- 30
- 60

poyribn:
I have crudely measured that the current minimum Framerate is 4 or 5 FpS.

Changing from 5 FpS to 3 FpS would increase the Time by ⅔.
Changing from 4 FpS to 3 FpS would increase the Time by ⅓.

Considering that this randomized Image saved almost instantly at the current minimum Setting, it would still save fast at 3 FpS:

<img width="683" height="384" alt="Image" src="https://github.com/user-attachments/assets/38cd2546-976f-443a-9376-6ed047a1c60c" />

poyribn:
Today i downloaded a 463 kB File from cimbar.org.
Using the lowest Framerate and the highest Framerate both took equally long,
so setting it to 3 FpS wouldn’t make a meaningful Difference in Speed.
But it would make a meaningful Difference in Safety.

sz3:
[...]

Issue #80, opened 2023-11-13
If you start libcimbar.org for I while you start seeing moving things and dots and cool patterns (cool thing I found not bug)


lnee94:

sz3:
It does have something of a hypnotic quality...

lnee94:
It is pure uninterpatable chaos which makes it prime for your brain to step in and add stuff. For example I stare my window blinds and see them moving I think this is the same effect.

PR #186, opened 2026-08-18
css cleanup (no "invisible-click" element) and seizure warning

sz3:
#180 points out that cimbar could pose some danger to those with photosensitive epilepsy. I'm a *bit* skeptical of this being a problem during normal operation, where you're probably not staring at a bright screen but mostly checking in on your phone screen (which is "staring" at the screen). OTOH, if you're staring into the pretty static (#80) and have photosensitivity, you should probably reconsider...

*Anyway*, 15 fps is, AFAIK, right in the danger range. So we'll have a warning. I tried to keep it short and simple because nobody actually reads paragraphs of warning text.

(also didn't want an extra warning dialogue box to click through)

+ this requires some minor CSS changes. I also removed the "invisible-click" element hack to toggle the menu. (more extensive CSS rewrite for better keyboard/accessibility navigation will come soon)

### DIFF
--- a/web/index.html
+++ b/web/index.html
@@ -98,6 +98,10 @@
       font: 1.5em serif;
     }
 
+    .drop-message .msg-long {
+      font: 1.0em serif;
+    }
+
     .drop-message .msg-bottom {
       margin-top: auto;
     }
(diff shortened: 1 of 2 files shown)


web/index.html
```diff
@@ -330,16 +359,21 @@
   </style>
 
   <div id="status"></div>
-  <div id="dragdrop" class="dragdrop" title="Drag and drop a file">
+  <div id="dragdrop" class="dragdrop" title="Drag and drop a file" onclick="Main.clickNav()">
+    <div class="drop-message">
+      <p class="msg-default">⌜ Tap to start! ⌟</p>
+      <p class="msg-error">⌜ Error! WebGL is not enabled :( ⌟</p>
+      <p class="msg-default msg-bottom">⚠️⚡ PHOTOSENSITIVITY WARNING!</p>
+      <p class="msg-default msg-long">Photosensitive epilepsy + flickering light = seizure!</p>
```
sz3, line 367: Short and to the point, hopefully

web/index.html
```diff
@@ -330,16 +359,21 @@
...
+      <p class="msg-default msg-long">Be safe! Have fun! ⚠️⚡</p>
```
sz3, line 368: With some words of encouragement. 🙂

